PR #186, opened 2017-01-13
Update license

mager:
If you use LICENSE.md instead, Github will display it in the header of the project.


blt:
Oh, hmm, neat idea but since the text is modified it won't auto-display correctly. I'm going to remove this modification.

PR #195, opened 2017-01-16
Flush interval is separated for every sink

### LINKED ISSUE #74: sinks' flush intervals ought to be independent
Right now cernan has a global concept of flush intervals. It would be much nicer if sinks had a per-sink flush interval, allowing us to differentiate between disk IO flushes and network flushes.

gliush:
Resolves #74

### DIFF
--- a/src/bin/cernan.rs
+++ b/src/bin/cernan.rs
@@ -73,11 +73,14 @@ fn main() {
 
     // SINKS
     //
-    let mut flush_sends = Vec::new();
     if let Some(config) = args.console {
         let (console_send, console_recv) =
             hopper::channel(&config.config_path, &args.data_directory).unwrap();
-        flush_sends.push(console_send.clone());
+        let flush_channel = console_send.clone();
+        let flush_interval = config.flush_interval;
+        joins.push(thread::spawn(move || {
+            cernan::source::FlushTimer::new(flush_channel, flush_interval).run();
+        }));
         sends.insert(config.config_path.clone(), console_send);
         joins.push(thread::spawn(move || {
             cernan::sink::Console::new(config).run(console_recv);
@@ -86,7 +89,11 @@ fn main() {
     if let Some(config) = args.null {
         let (null_send, null_recv) = hopper::channel(&config.config_path, &args.data_directory)
             .unwrap();
-        flush_sends.push(null_send.clone());
+        let flush_channel = null_send.clone();
+        let flush_interval = config.flush_interval;
+        joins.push(thread::spawn(move || {
+            cernan::source::FlushTimer::new(flush_channel, flush_interval).run();
+        }));
         sends.insert(config.config_path.clone(), null_send);
         joins.push(thread::spawn(move || {
             cernan::sink::Null::new(config).run(null_recv);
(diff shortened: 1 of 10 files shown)


blt:
Reviewed 4 of 10 files at r1.
Review status: 4 of 10 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/config.rs, line 33 at r1](https://reviewable.io:443/reviews/postmates/cernan/195#-KadTtlis7wzPi039Xs6:-KadTtlis7wzPi039Xs7:b-hvdcwb) ([raw file](https://github.com/postmates/cernan/blob/e5fed1453ee5030bc6a1366e0223256f60a826df/src/config.rs#L33)):*
> ```Rust
>     pub filters: HashMap<String, ProgrammableFilterConfig>,
>     pub firehosen: Vec<FirehoseConfig>,
>     pub default_flush_interval: u64,
> ```

It's not documented anywhere but in my mind I've kept a distinction between "global" and "local" settings. Could you rename this to `global_flush_interval`. Then, optionally, a sink may provide a "local" `flush_interval`. 

---

*[src/config.rs, line 82 at r1](https://reviewable.io:443/reviews/postmates/cernan/195#-KadUhilK0TR_6hub7-8:-KadUhilK0TR_6hub7-9:ba5d0wt) ([raw file](https://github.com/postmates/cernan/blob/e5fed1453ee5030bc6a1366e0223256f60a826df/src/config.rs#L82)):*
> ```Rust
>         // We read from CLI arguments
>         None => {
>             let default_flush_interval = u64::from_str(args.value_of("default-flush-interval").unwrap())
> ```

I don't want to add anymore flags to the CLI. Let's remove this. 

---

*[src/config.rs, line 91 at r1](https://reviewable.io:443/reviews/postmates/cernan/195#-KadUqS_QSi6OBce6Y-G:-KadUqS_QSi6OBce6Y-H:b-w2qoyv) ([raw file](https://github.com/postmates/cernan/blob/e5fed1453ee5030bc6a1366e0223256f60a826df/src/config.rs#L91)):*
> ```Rust
>                     config_path: "sinks.wavefront".to_string(),
>                     tags: Default::default(),
>                     flush_interval: args.value_of("wavefront-flush-interval")
> ```

It's acceptable to me to set this to `60`. The CLI interface is from ancient-cernan and we need to think through how it ought to look if we do want to support it. 

---

*[src/bin/cernan.rs, line 206 at r1](https://reviewable.io:443/reviews/postmates/cernan/195#-KadecPktuw0HKi2I2PX:-KadecPktuw0HKi2I2PY:b-92lwgh) ([raw file](https://github.com/postmates/cernan/blob/af1287323b6b35fe1cc48ae0c49a073fdcbe11a4/src/bin/cernan.rs#L206)):*
> ```Rust
>     //
> 
>     let flush_interval = args.flush_interval;
> ```

Rather than create a new `FlushTimer` for every enabled sink can you work it so that there's still one `FlushTimer` but it services all the flush emissions? Probably something like `[(sleep_time, [snd_channels])]` to make it work out. I'm already a little hesitant concerning the number of OS threads cernan makes and I'd like to avoid making more if we don't have to. 

---

*[src/bin/cernan.rs, line 207 at r1](https://reviewable.io:443/reviews/postmates/cernan/195#-KadgDEg_WqTMIsEouUA:-KadgDEg_WqTMIsEouUB:b-y6jdpf) ([raw file](https://github.com/postmates/cernan/blob/af1287323b6b35fe1cc48ae0c49a073fdcbe11a4/src/bin/cernan.rs#L207)):*
> ```Rust
> 
>     let flush_interval = args.flush_interval;
>     joins.push(thread::spawn(move || {
> ```

Oh crap, we might not be able to split the 

---

*[src/sink/null.rs, line 21 at r1](https://reviewable.io:443/reviews/postmates/cernan/195#-KadfzX_lAH_STmm5xZu:-KadfzX_lAH_STmm5xZv:b-56645b) ([raw file](https://github.com/postmates/cernan/blob/e5fed1453ee5030bc6a1366e0223256f60a826df/src/sink/null.rs#L21)):*
> ```Rust
> 
> impl NullConfig {
>     pub fn new(config_path: String, flush_interval: u64) -> NullConfig {
> ```

Because the null sink does nothing I'm satisfied not providing it with a local `flush_interval`. It can be driven off the global interval. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/postmates/cernan/195)*

blt:
Oh dear, oh no, those non-inline review comments are not good. Nope.

blt:
@gliush Oh hmm, waaaait a second. Unless I'm reading the master branch incorrectly there's a defect here.  Every modification of [`flush_sends`](https://github.com/postmates/cernan/blob/master/src/bin/cernan.rs#L76) is populating _only at the base_ of the topology. This means that flushes are _not_ getting sent through filters like we intended. Crap! 

I'm going to create a bug for this.

gliush:
Outdated

PR #198, opened 2017-01-25
counter and gauge correct sample syntax

gliush:


blt:
When you smash this for merging if you could write a complete commit message with subject and body I'd appreciate it. Something so that when I'm putting together release notes I know which issue this is related to / why the work was done. Previous commits in-project are an example of what I mean.

blt:
Going to do this myself. Thanks @gliush! :D

PR #200, opened 2017-01-26
Get travis to check the format of branches

blt:
This PR introduces running `cargo fmt -- --write-mode=diff` against every branch to ensure that style is maintained. This _may_ be annoying as differing versions of rustfmt seem to do different text manipulations but we'll see. If it is annoying we can always back it out.


gliush:
LGTM

PR #204, opened 2017-02-03
Update cernan to use serde 0.9

blt:
This commit removes the special codegen step that we had to take with
rust < 1.15 and serde < 0.9. Now that 1.15 is Stable we can remove
this codegen step. Builds take about the same speed but the code
is a little nicer to interact with.


blt:
@tsantero thanks!

PR #206, opened 2017-02-15
Flush intervals rework

gliush:
Flush intervals rework:
- Now flushes go every second and contain `flush index` value (u32)
- They pass through filters as well
- If a flush comes to the sink by different ways (from source + from filter[s]), only the first flush with the same `flush index` will be handled, all the next will be skipped

### DIFF
--- a/src/bin/cernan.rs
+++ b/src/bin/cernan.rs
@@ -189,7 +189,7 @@ fn main() {
     // BACKGROUND
     //
 
-    let flush_interval = args.flush_interval;
+    let flush_interval = 1; // TODO: use constant
     joins.push(thread::spawn(move || {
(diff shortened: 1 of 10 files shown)


blt:
Does this supplant #195? If so, can we close that PR out?

src/bin/cernan.rs
```diff
@@ -189,7 +189,7 @@ fn main() {
     // BACKGROUND
     //
 
-    let flush_interval = args.flush_interval;
+    let flush_interval = 1; // TODO: use constant
```
blt: TODO left. That said, the optimizer should detect that this is a constant so I'm OK with not introducing some global constant.

gliush: removed TODO

gliush:
Outdated

PR #213, opened 2017-02-21
Flush interval3

gliush:
It's a new attempt to implement a flush intervals.
Previous attempts and discussions see here: [Part1](https://github.com/postmates/cernan/pull/195), [Part2](https://github.com/postmates/cernan/pull/206)

No tests yet (not clear how to test this functionality)


blt:
@doubleyou once this is merged you'll have the ability to declare the global flush interval to be one second and individual sinks to sixty seconds. That way the filters you're hoping to write will get a tick once-per-second – allowing you to do rates and what not – without having to simultaneously increase the rate of transmission to the sinks.

blt:
@doubleyou scratch that last somewhat. As of 8a5edf7 the filter `tick` will be guaranteed to be called once per second. The global `flush-interval` will now act as a default value for sinks that don't otherwise set `flush_interval`.

blt:
@tsantero @gliush could you both take a look at this PR? I think it's ready to go.

gliush:
It looks ok for me.
But as we discussed previously, we should have tests for this functionality, and to do it now - is very tricky.
I've already done refactoring of `main`, so that it will be easier (will send email shortly) so we might wait another week to go through all the reviews and release.

PR #220, opened 2017-03-05
Introduce payload clearing functions

blt:
In conversation with @doubleyou I've learned that there's a use-case
for a filter that reads in a metric and then blanks the contents of
the payload. This is especially useful for filters that compute
analytics but want to dead-end streams of incoming metrics
excepting those that have been computed.

### DIFF
--- a/src/filter/programmable_filter.rs
+++ b/src/filter/programmable_filter.rs
@@ -89,7 +89,7 @@ impl<'a> Payload<'a> {
     }
 
     #[allow(non_snake_case)]
-    unsafe extern "C" fn lua_clear_metric(L: *mut lua_State) -> c_int {
+    unsafe extern "C" fn lua_clear_metrics(L: *mut lua_State) -> c_int {
         let mut state = State::from_ptr(L);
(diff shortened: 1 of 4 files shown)


src/filter/programmable_filter.rs
```diff
@@ -300,19 +315,21 @@ impl<'a> Payload<'a> {
     }
 }
 
-const PAYLOAD_LIB: [(&'static str, Function); 12] =
-    [("metric_name", Some(Payload::lua_metric_name)),
-     ("metric_query", Some(Payload::lua_metric_query)),
+const PAYLOAD_LIB: [(&'static str, Function); 14] =
+    [("set_metric_name", Some(Payload::lua_set_metric_name)),
+     ("clear_logs", Some(Payload::lua_clear_log)),
+     ("clear_metrics", Some(Payload::lua_clear_metric)),
```
pulltab: Why isn't this called lua_clear_metrics?

blt: Because I accidentally broke naming conventions like a goof. :) 

Will fix.

Issue #221, opened 2017-03-05
Consider moving configuration format from TOML to YAML 


blt:
Since cernan introduced a configuration file it's used the TOML format, common in rust. For cernan's purpose there are some potential defects to the use of TOML. Namely:

  * As cernan's configuration file grows in complexity we hit up against TOML's limitations. #218 runs up against this. 
  * There is no inclusion capability, making cernan's config a one-shot deal. 
  * Interaction with the TOML library is awkward. config.rs is the largest handwritten file in the project.

That's two points against the format, one point against the library we use. Ultimately the final point doesn't matter so long as we can parse something that is effective for end-users and give valuable feedback on parsing errors.

blt:
YAML is whitespace sensitive. Do not want.

PR #222, opened 2017-03-05
Introduce configurable percentiles for wavefront sink

### LINKED ISSUE #218: Make WF percentiles configurable
Right now we pump out hard-coded percentiles from the wavefront sink. Many of them are likely not very useful and we should make these percentiles configurable.

blt:
This commit makes it possible for the end-user to specify what
percentiles will be shipped out in the wavefront sink. The config
for doing so is... goofy. This problem is discussed in #221.

This commit resolves #218.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -80,11 +80,25 @@ pub fn parse_args() -> Args {
         // We read from CLI arguments
         None => {
             let wavefront = if args.is_present("wavefront") {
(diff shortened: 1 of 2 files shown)


gliush:
LGTM, though, I don't like that we have the same percentiles initialisation code (though, I realise that right now we can't fix it).

PR #223, opened 2017-03-05
Main function refactor

gliush:
Hi.

I reworked the main function (DRY principal + some simplifications).
Though, I still don't know how to avoid:
* `Entry:run1` function, that I had to implement in all the modules. It's almost the same as Sink:run | Source:run | Filter:run. Ideally, I'd like to write a default implementation in {sink,source,filter}/mod.rs, but I couldn't. I even not sure that it's doable.

I'm waiting for a review and suggestions!

### DIFF
--- a/src/entry/mod.rs
+++ b/src/entry/mod.rs
@@ -1,0 +1,15 @@
+use hopper;
+use metric::Event;
+
+pub trait Entry {
+    // fn run();
+    fn get_config(&self) -> &EntryConfig;
+    fn run1(&mut self, _forwards: Vec<hopper::Sender<Event>>, recv: hopper::Receiver<Event>);
+}
+
+pub trait EntryConfig {
+    fn get_forwards(&self) -> Vec<String> {
+        Vec::new()
+    }
+    fn get_config_path(&self) -> &String;
+}
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ target/
 src/metrics/statsd.rs
 src/metrics/graphite.rs
 *.bk
+/tags
+/.vscode
--- a/src/bin/cernan.rs
+++ b/src/bin/cernan.rs
@@ -6,7 +6,7 @@ extern crate log;
 extern crate hopper;
 
 use cernan::entry::Entry;
-use cernan::filter::{Filter, ProgrammableFilterConfig};
+use cernan::filter::ProgrammableFilterConfig;
 use cernan::metric;
 use cernan::sink;
 use cernan::source::Source;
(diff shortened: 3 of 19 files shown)


src/entry/mod.rs
```diff
@@ -0,0 +1,15 @@
+use hopper;
+use metric::Event;
+
+pub trait Entry {
```
blt: It occurs to me that the `Entry` concept has a conceptual problem, I think. We don't really _need_ the full configuration as received from the users, only the bits that are important to the operation of the sink. That makes the ownership model of the config straightforward: we move the pieces of use into the sink and drop the rest. `Entry` retains ownership of the config – necessitating cloning – but I'm not sure _why_ this is needed. `Entry` finds use in cernan.rs for building up the forwards but the implementation is less straightforward than the approach in #213. Testing #213 is hard and maybe an explicit `Entry` helps with that? 

Also, keeping `Entry` as such in memory is problematic for credentials, especially those systems where it's possible to compute a session key from long-lived credentials. Right now we've got a design that'll allow us to dump those long-lived credentials through deallocation but `Entry` – I think – would lock us into keeping them in-memory for the lifetime of the cernan instance.

gliush: Actually, we don't need the full config, only "forwarding" sections. So I should simplify "Entry" to contain only copy of it.

About simplifications in testing. 
In the `master` the approach is to go all the sinks/sources/filters one by one, creating the threads. So we can't test it.
In my approach I've packed it into two separate cycles.
Yes, it is still impossible to test, but it is only a first step.

The next step would be to have 3 cycles: 
- to initialyze data
- given graphs of "followers"  make some conclusions 
- run threads.
If the first two cycles are in function - it makes it very easy to test.
I should have done it in this PR, I agree, so I'll do it here.

src/entry/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+    // fn run();
+    fn get_config(&self) -> &EntryConfig;
```
gliush: That's for sure!

src/entry/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+    fn run1(&mut self, _forwards: Vec<hopper::Sender<Event>>, recv: hopper::Receiver<Event>);
```
gliush: Well, this is the reason why I made a PR so early.
I don't know how to structure the code so that I didn't have this function.
I mean, the only thing that `run1` does - it calls `run`.
But as the Trait is defined for different types (Sink, Source, Filter) - I can't write a proper trait that will ensure that the function exists.
To rephrase: 
I need a new trait (`Entry` in my code) that have function (`run1`), such that this function would start all types of objects (`Sink`, `Source`, `Filter`).
We already have `run` for `Sink`, `Source`, `Filter`, so I can't make `Entry` have `run`. This is why I defined `run1`, and now I have to write implementation for  it in all sinks, in all sources and filters. Though, it has the same implementation `fn run1() -> run() end` (in pseudocode).
I hope that I missed something in language syntax so that I have trait `Entry` with default `run1` function. Is it possible?

blt:
@gliush once #259 lands I'm going to close this PR if you've got no objections.

gliush:
Sure.

Issue #229, opened 2017-03-10
Support JSON as a first class concern


blt:
Multiple folks have asked for the filter subsystem to be able to encode and decode JSON. It's possible to slurp a JSON decoder into your filter but that's not viable long-term. cernan should provide an API for manipulating JSON. We can almost surely just re-expose serde.

blt:
The test in #241 demonstrates the method of including pure-lua json parsers. This issue is now closed! :D

PR #230, opened 2017-03-17
Allow lua filters to require other lua modules

blt:
This commit add explicit support for requiring modules in cernan
programmable filters, pegged by default to script-directory. We
make no attempts to disallow further modification of package.path
and the pegging to script-directory is meant to give the filter
programmer access to said directory without needing to know
exactly how cernan has been configured.

This is part of a larger work to support JSON as an input to filters.
See #229 for more details.

### DIFF
--- a/resources/tests/scripts/demonstrate_require.lua
+++ b/resources/tests/scripts/demonstrate_require.lua
@@ -1,0 +1,11 @@
+local demo = require "lib/demo"
+
+function process_metric(pyld)
+end
+
+function process_log(pyld)
+   payload.log_set_tag(pyld, 1, "bizz", demo.demo())
+end
+
+function tick(pyld)
+end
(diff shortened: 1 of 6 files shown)


resources/tests/scripts/demonstrate_require.lua
```diff
@@ -0,0 +1,11 @@
+local demo = require "lib/demo"
```
pulltab: The goal here is to require a package installed via luarocks or somesuch?

If so, can we add a test case that covers requiring a globally installed lua dependency?

blt: The idea is that you ought to be able to include other files in your scripts directory. I only had ambitions for native lua files and haven't tested mixed lua/C setups. 

I restricted the path to _just_ scripts-directory, though it's still possible for the filter programmer to change the path to be global. That is, I am _assuming_ local includes but not denying global, if the user goes to the trouble of modifying the restricted path appropriately.

PR #234, opened 2017-03-29
Increase the responsiveness of the file source

### LINKED ISSUE #233: Cope with fast/slow combination of logs in a single glob
In conversation with @lynchc we discovered that if you combine fast and slow logs under a single file glob then there will be long periods where the fast log will not be addressed. This a complication of our present algorithm where all files are given the same priority and delays are introduced between file reads, crossed with caps on total lines read per interval. 

The intention for the file source is to be "fair". That is, in the 60 second time-slice available to cernan absorb as many lines as possible _without_ polling continuously – which blows out a CPU core – but without also losing lines to rotation.

blt:
Consider a situation where there are a few very fast log files but
mostly slow ones. What happens? Well, previously we waited 512ms
between files _without_ reducing that wait. Now we zip the wait down
to 0ms if we've hit a file that has changes, which _should_ improve
responsiveness in mixed-speed situations.

Additionally the file source has grown a max_read_lines configuration
option, allowing operators to decide how many lines to read from an
individual file before moving on to the next. By default this value
is 10k.

This resolves #233

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -383,11 +383,26 @@ pub fn parse_config_file(buffer: String, verbosity: u64) -> Args {
                             }
                             None => Vec::new(),
                         };
+                        // NOTE The table lookup will return an i64 but we
+                        // convert to usize. Strictly speaking this IS NOT a
+                        // safe conversion but it makes no good sense for a user
+                        // to be reading anywhere above 2**60 lines per file. We
+                        // leave these people to their own wild works and hope
+                        // for the best.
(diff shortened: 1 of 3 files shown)


gliush:
LGTM
`carma+=1` for comments! :)

blt:
Hmm, so the codecov/patch coverage is 90% in the report _but_ I don't see in their UI that coverage has dropped anywhere. Rather, it's gone up.

PR #235, opened 2017-03-29
Introduce the 'internal' source

### LINKED ISSUE #212: Introduce 'internal' source
Cernan very usefully produces self-telemetry. `cernan.statsd.bad_packets` is one such example. The amount of self-telemetry is fairly low. At present a component of cernan must create a new Telemetry instance, feed it through the configured pipeline and then out. This is problematic for a handful of reasons: 

* operators are unable to emit self-telemetry along a different topology,
* sinks cannot 'go up' the topology and must emit self-telemetry to their far target and
* the lower in the routing topology you are, the less introspection capability. 

If we introduce a new kind of source that obeys routing rules but exposes no external interface we can fix these problems. Call it "internal". The 'internal' source should provide some API to components such that they can create Telemetry as needed but still be hooked up to the queue mechanism for routing. That way cernan components can self-telemeter with abandon and operators can route it all as usual.

blt:
This new source allows cernan to self-telemeter. It is not possible
for modules to report telemetry 'up' to be distributed through
the filter topology as operators see fit. This interface is
intentionally minimal and can be extended as need arises.

Closes #212

### DIFF
--- a/src/source/internal.rs
+++ b/src/source/internal.rs
@@ -12,7 +12,7 @@ lazy_static! {
 /// 'Internal' is a Source which is meant to allow cernan to
 /// self-telemeter. This is an improvement over past methods as an explicit
 /// Source gives operators the ability to define a filter topology for such
-/// telemetry and makes it easier to modules to report on themeselves. 
+/// telemetry and makes it easier for modules to report on themeselves.
 pub struct Internal {
     chans: util::Channel,
     tags: sync::Arc<metric::TagMap>,
@@ -25,7 +25,8 @@ pub struct InternalConfig {
     pub config_path: String,
     /// The forwards which Internal will obey.
     pub forwards: Vec<String>,
-    /// The default tags to apply to each Telemetry that comes through the queue.
+    /// The default tags to apply to each Telemetry that comes through the
+    /// queue.
     pub tags: metric::TagMap,
 }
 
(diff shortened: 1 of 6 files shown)


src/source/statsd.rs
```diff
@@ -65,14 +66,10 @@ fn handle_udp(mut chans: util::Channel, tags: sync::Arc<metric::TagMap>, socket:
                     for m in metrics.drain(..) {
                         send("statsd", &mut chans, metric::Event::new_telemetry(m));
                     }
-                    let mut metric = metric::Telemetry::new("cernan.statsd.packet", 1.0).aggr_sum();
-                    metric = metric.overlay_tags_from_map(&tags);
-                    send("statsd", &mut chans, metric::Event::new_telemetry(metric));
+
```
pulltab: Whitespace police.

src/source/internal.rs
```diff
@@ -0,0 +1,90 @@
...
+use util;
+
+lazy_static! {
+    static ref Q: sync::Mutex<VecDeque<metric::Telemetry>> = sync::Mutex::new(VecDeque::new());
+}
+
+/// 'Internal' is a Source which is meant to allow cernan to
+/// self-telemeter. This is an improvement over past methods as an explicit
+/// Source gives operators the ability to define a filter topology for such
+/// telemetry and makes it easier to modules to report on themeselves.
```
pulltab: *for modules to

PR #237, opened 2017-03-30
Increase existing telemetry, adjust file delays

blt:
This commit contains a smattering of adjustments. The main show
is the inclusion of Internal setup code in bin/cernan -- oops, it
was only present on my disk -- and fiddling with the way backoffs
occur in file.rs. We reduce the CPU load of each FileServer thread
by only gradually scaling down to faster polls. This means a group
of slow files will be polled more slowly that a group with one
fast file in it.

### DIFF
--- a/src/source/mod.rs
+++ b/src/source/mod.rs
@@ -8,7 +8,7 @@ mod statsd;
 pub use self::file::{FileServer, FileServerConfig};
 pub use self::flush::FlushTimer;
 pub use self::graphite::{Graphite, GraphiteConfig};
-pub use self::internal::{Internal, InternalConfig};
+pub use self::internal::{Internal, InternalConfig, report_telemetry};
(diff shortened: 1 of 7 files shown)


src/source/mod.rs
```diff
@@ -8,7 +8,7 @@ mod statsd;
 pub use self::file::{FileServer, FileServerConfig};
 pub use self::flush::FlushTimer;
 pub use self::graphite::{Graphite, GraphiteConfig};
-pub use self::internal::{Internal, InternalConfig};
+pub use self::internal::{Internal, InternalConfig, report_telemetry};
```
pulltab: Where does report_telemetry come from?

blt: @pulltab it is defined in internal.rs, here: https://github.com/postmates/cernan/blob/master/src/source/internal.rs#L57

Issue #205, opened 2017-02-06
First build fails when doing a static build using build-container.sh


blakebarnett:
```
musl-gcc -O2 -Wall -Wextra -DLUA_COMPAT_5_2 -DLUA_USE_LINUX -O3 -ffunction-sections -fdata-sections -fPIC -m64 -static    -c -o lua.o /home/rust/.cargo/registry/src/github.com-1ecc6299db9ec823/lua-0.0.10/lua-source/src/lua.c
<builtin>: recipe for target 'lua.o' failed
make[1]: Leaving directory '/home/rust/src/target/x86_64-unknown-linux-musl/release/build/lua-dcf49d3524bf0676/out'
/home/rust/.cargo/registry/src/github.com-1ecc6299db9ec823/lua-0.0.10/lua-source/src/Makefile:110: recipe for target 'linux' failed

--- stderr
ar: `u' modifier ignored since `D' is the default (see `U')
/home/rust/.cargo/registry/src/github.com-1ecc6299db9ec823/lua-0.0.10/lua-source/src/lua.c:80:31: fatal error: readline/readline.h: No such file or directory
compilation terminated.
make[1]: *** [lua.o] Error 1
make: *** [linux] Error 2
```

Subsequent builds work fine, it seems to be some kind of race condition building lua.

blt:
Resolved by #238

PR #238, opened 2017-04-03
Update docker build

blt:
This PR introduces updates to the image building process for cernan. The ambition is to build an image for each new release and in-progress cuts of master as well. The busybox:glibc image is used as a base for production -- see Dockerfile.prod -- and postmates/postmates-rust.

The image built in this commit is not quite correct but it's a start. It also depends on closed-source postmates-rust, a thing which we will need to open-source.


blt:
This resolves #205

blt:
I think that's a good idea. "basic" got pretty bloated over time.

Regarding the curl and OpenSSL needs, I think I have a line on that.

tsantero:
@blt can we discuss what you've done here sometime soon? i have very similar needs

PR #239, opened 2017-04-03
Cope with prometheus header behaviour

blt:
It turns out that prometheus is a little more lax in practice with
regard to content-type headers than its documentation promises. This
commit meets prometheus where it is, emitting text/plain if there's
no header and not emitting protobuf by default.

### DIFF
--- a/src/sink/prometheus.rs
+++ b/src/sink/prometheus.rs
@@ -88,10 +88,18 @@ fn write_text(aggrs: AggrMap, mut res: Response) {
     let mut buf = String::with_capacity(1024);
     let mut res = res.start().unwrap();
     for m in aggrs.into_iter() {
+        let sum_tags = m.tags.clone();
+        let count_tags = m.tags.clone();
         for q in &[0.0, 1.0, 0.25, 0.5, 0.75, 0.90, 0.95, 0.99, 0.999] {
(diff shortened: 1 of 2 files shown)


src/sink/prometheus.rs
```diff
@@ -118,14 +120,34 @@ impl Handler for SenderHandler {
...
+        let raw_headers: Vec<&str> = req.headers
+            .get_raw("content-type")
+            .unwrap_or(&[])
+            .iter()
+            .map(|x| str::from_utf8(x))
+            .filter(|x| x.is_ok())
+            .map(|x| x.unwrap())
+            .collect();
+        assert!(raw_headers.len() <= 1);
+        if raw_headers.is_empty() {
```
pulltab: Do we have unit test coverage for both of these branches?

blt: We don't. Apparently Past Brian was a bum and there are no tests for this module. 

I will correct that.

pulltab: Per discussion with blt, Prometheus is difficult to unit test against and better integration tests are planned once an improved CI pipeline is in place.

blt:
The patch failure does not appear to be accurate. It's dinging function
headers, which have to run, and comments, which cannot run.

PR #240, opened 2017-04-04
Remove all but the median and 99th percentile

bitglue:
Wavefront costs us about $20,000 per month, and growing. Shipping fewer
percentiles is a quick and easy experiment to determine:

1. What fraction of our metrics are timer percentiles?
2. Is anyone using these?

It may end up being a significant reduction in operating expense.

INFRASTRUC-3486


bitglue:
:frowning_face:

PR #241, opened 2017-04-05
Add 'fields' to LogLine, filter manipulation of such

blt:
We need to start parsing structured information out of LogLines and this
work enables it. What we've done is plonk an additional k/v structure
inside the LogLine that filters have access to, are able to manipulate.
We demonstrate this under test by pulling information from a json payload.


blt:
This resolves #229

PR #242, opened 2017-04-09
Build a special-purpose aggregator for Prometheus

blt:
The constraints placed by Prometheus on a reporter are non-trivial,
especially with regard to reports through time. This commit introduces
a new aggregator for Prometheus and a battery of quickcheck tests to
demonstrate singleton reports of time-series / timestamp, re-introduction
of telemetry after report failure and documentation of such.

We have to discover Prometheus' constraints, in part, and the model
presented in this commit may diverge from Prometheus' actual demands.

### DIFF
--- a/src/metric/mod.rs
+++ b/src/metric/mod.rs
@@ -1,4 +1,4 @@
-mod tagmap;
+pub mod tagmap;
 mod logline;
 mod event;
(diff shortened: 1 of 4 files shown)


blt:
@blakebarnett relevant to your interests

PR #251, opened 2017-04-12
Introduce 'mean' into wavefront summarize reporting

### LINKED ISSUE #246: Timers don't report the mean
Quantiles are good, but sometimes you just want an ordinary arithmetic mean. Also necessary for feature parity with most other statsd-like aggregators.

blt:
As of quantiles 0.4.0 it's possible to get the mean of a quantile
structure. Telemetry is now aware of its mean and we expose this
into the wavefront report, along with count.

This resolves #246

### DIFF
--- a/src/sink/wavefront.rs
+++ b/src/sink/wavefront.rs
@@ -147,6 +147,19 @@ impl Wavefront {
                         self.stats.push_str(&tag_buf);
                         self.stats.push_str("\n");
 
+                        let mean = value.mean();
+                        self.stats.push_str(&value.name);
+                        self.stats.push_str(".mean");
+                        self.stats.push_str(" ");
+                        self.stats
+                            .push_str(get_from_cache(&mut value_cache, mean));
+                        self.stats.push_str(" ");
+                        self.stats
+                            .push_str(get_from_cache(&mut time_cache, value.timestamp));
+                        self.stats.push_str(" ");
+                        self.stats.push_str(&tag_buf);
+                        self.stats.push_str("\n");
+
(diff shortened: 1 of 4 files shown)


src/sink/wavefront.rs
```diff
@@ -335,6 +348,7 @@ mod test {
         assert!(lines.contains(&"test.timer.99 12.101 645181811 source=test-src"));
         assert!(lines.contains(&"test.timer.999 12.101 645181811 source=test-src"));
         assert!(lines.contains(&"test.timer.count 3 645181811 source=test-src"));
+        assert!(lines.contains(&"test.timer.mean 5.434333333333334 645181811 source=test-src"));
```
pulltab: What is the expected behavior when a 64 bit integer is passed in?

pulltab: 1) Should we test this expected behavior?

2) We should probably call out to the users that certain integer types are not supported.

blt: It will fail to compile because of the constraints placed on the CKMS type `T`. It's a pain, but at least you can't sleep-walk into an accident.

blt: Oh wait, maybe I'm misunderstanding your comments. The CKMS type in cernan is specialized to `CKMS<f64>` and everything is coded to that demand. It's only now that we've updated to quantiles 0.4.0 that `CKMS<T>` must be `T: Into<f64>`. The impact on cernan is nil.

pulltab: That's what I wanted to hear.

PR #258, opened 2017-05-01
Explicitly add ca-certificates to the release image

blt:
It occurs to me that maybe the firehose issue we're seeing is a
lack of certs sitting inside the release image. If this commit
finds its way to the master branch this was the case.


blakebarnett:
I'm pretty sure I tried this at some point, but hey... that'd be great if it works.

PR #263, opened 2017-05-22
Fix the name mangling issue with self-telemetry

### LINKED ISSUE #262: Self-telemetry plays hell with prometheus sink
@benley reports: 

> argh the metrics that cernan exports to prometheus are completely useless because they don't use labels
>
> Instead of names like `cernan_sources_file__var_log_containers_calico_node_0q8hz_kube_system_calico_node_c7610bc0d4e2853ec3a8feb81ae69f18833cfd43ede88720ad6297eb335f1f29_log_lines_read_count` it would be preferable to have a metric called `cernan_sources_log_lines_read_count{file="/var/log/containers/blahblahblahlongname"}`
>
> i.e. rather than constructing a separate metric name for each thing, use metric labels and combine them

Now, the prometheus _sink_ correctly sets labels. What's happening is that the file source uses `report_telemetry` to ship that metric. `report_telemetry` does allow the caller to set metadata which means that callers resort to cramming info into the telemetry name. 

This can be corrected by extending `report_telemetry`. Unfortunately, it will means that the internal metrics will be renamed.

blt:
When the internal metric source was created we implicitly encoded
a graphite-like scheme for metric naming, which, since cernan intends
to be agnostic to sink, was not ideal. This commit corrects that
mistake, allowing callers to report_telemetry to expand their
use to report_full_telemetry, a function which admits metadata and
changing the aggregation type.

This resolves #262

### DIFF
--- a/src/sink/firehose.rs
+++ b/src/sink/firehose.rs
@@ -12,7 +12,7 @@ use serde_json;
 use serde_json::Map;
 use serde_json::value::Value;
 use sink::{Sink, Valve};
-use source::report_telemetry;
+use source::report_full_telemetry;
 use std::sync;
 use uuid::Uuid;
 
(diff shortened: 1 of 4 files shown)


blt:
@postmates/infra FYI if this gets merged in it _will_ change cernan-internal metric names. Additionally, this may cause issues with downstream aggregators that have issues with tag cardinality. https://github.com/postmates/cernan/pull/263/files#diff-5dcbfff2f177e3df0aee469f38d03444R127, for example, will create one time series per file – which was already the case – but will do so now by sharing a common metric name plus a distinct set of tags for that common metric name.

blakebarnett:
This will effect anyone using the lua filters right?  Anything else you know of that will break?

bitglue:
Is this going to result in Wavefront metrics with a large number of tags? That will kill WF performance: they can't filter on tags in their database. They have to load the values for all the tags, then filter them client-side.

blt:
@blakebarnett the lua filters ought to be alright. `file` source and `firehose` sink will have their metric names change. All other metrics – application and internal metrics not from the above sources, sinks – will remain unchanged. Also, @bitglue is correct. This will tank performance for cernan's internal `file` and `firehose` metrics in Wavefront. I'm unsure if increasing the cardinality on some tags changes performance characteristics across all time series or not. A lua filter _could_ be written to preserve the 0.5.x behaviour by stripping tags and modifying key metric names. 

The upside of this is that prometheus will be useful with regard to the internal metrics above.

bitglue:
My understanding is it impacts only the metric name which has a large number of tags.

Issue #224, opened 2017-03-07
DNS hostnames should not be blank


blt:
We have a crash report from @pulltab where a blank DNS hostname for the wavefront sink results in the following log messages: 

```
[cernan::sink::wavefront][182][2017-03-07T18:46:01.616903074+00:00][INFO] Unable to perform DNS lookup on host  with error failed to lookup address information: Name or service not known
[cernan::sink::wavefront][182][2017-03-07T18:46:01.616952296+00:00][INFO] Unable to perform DNS lookup on host  with error failed to lookup address information: Name or service not known
```

The config to cause this is:

```
[sinks]
  [sinks.wavefront]
  host=""
  port=2878
```

What cernan _ought_ to do is refuse to start if any of its hostnames are blank. They're simply not valid.

PR #265, opened 2017-06-03
Add minimal error handling for empty host names for wavefront

tfheen:
Instead of later panicking, make Wavefront::new return a Result and
match on that.

Fixes: https://github.com/postmates/cernan/issues/224

### DIFF
--- a/src/bin/cernan.rs
+++ b/src/bin/cernan.rs
@@ -105,8 +105,16 @@ fn main() {
             .unwrap();
         sends.insert(config_path.clone(), wf_send);
         joins.push(thread::spawn(move || {
-                                     cernan::sink::Wavefront::new(config).run(wf_recv);
-                                 }));
+            match cernan::sink::Wavefront::new(config) {
+                Ok(mut w) => {
+                    w.run(wf_recv);
+                }
+                Err(e) => {
+                    error!("Configuration error for Wavefront: {}", e);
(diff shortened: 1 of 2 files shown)


tfheen:
Yeah, I agree about that it should be fixed elsewhere too, but I'm still in "getting my feet wet with Rust" mode and wanted to see that I was on the right track before I started looking at bigger or harder problems.

Weird about the compilation problems, I'll reproduce and fix.

blt:
> Yeah, I agree about that it should be fixed elsewhere too, but I'm still in "getting my feet wet with Rust" mode and wanted to see that I was on the right track before I started looking at bigger or harder problems.

Absolutely. Just to be very clear, I don't think you need to tackle any of that in this PR (or at all, unless you eventually feel like it). More, I was just making note of it for myself and the other folks that watch this project.

tfheen:
Oh, another question: Do you care about Signed-off-By lines? I see you're using them, but I tend not to (unless the project wants it). And do you want rebases with fixed commits, or do you prefer a series of commits?

blt:
@tfheen thanks for asking! No Signed-off By lines are fine. I stick them in by default but I'm not a stickler about it. A series of commits is also fine. This project squashes all commits into a PR before merging into master.

tfheen:
Thanks for the review and feedback (and the merge)!

PR #268, opened 2017-06-09
Direct elasticsearch integration

blt:
This PR adds direct elasticsearch integration into cernan. This means that it's not longer necessary to go through AWS Firehose for dropping logs into ES. Please see individual commits for details.


src/sink/elasticsearch.rs
```diff
@@ -65,6 +66,54 @@ impl Elasticsearch {
...
+                Ok(())
+            }
+            // The index exists, but the doc wasn't found: map and index
+            Ok(_) => self.put_doc(idx.clone(), doc).map(|_| ()),
+            // No index: create it, then map and index
+            Err(error::Error(error::ErrorKind::Api(error::ApiError::IndexNotFound {
+                                                         ..
+                                                     }),
+                                      _)) => {
+                self.put_index(idx.clone()).and(self.put_doc(idx, doc)).map(|_| ())
```
tfheen: Isn't there a race condition here if you have multiple cernans pushing to the same ES server and there's an index change (say you're pushing to logs-%y%m%d and you can hit this around midnight)?

blt: I'm not totally satisfied with the error handling here. You're correct, it's possible that two simultaneous writers could step on each other. I need to learn more about Elasticsearch's write model to learn what kind of error it is. 

If in fact this does result in an error, the losing write will get pushed to the back of the buffer and re-tried at a later date. Not really ideal if it can be helped. There should also be documentation in this module / around this concern.

tfheen: Good point, as long as it's retried and eventually consistent, it should be fine, the particular error I'm thinking about should only happen once per instance per index switch (at worst), so it's not like it should be a problem in practice.

blt: We've got a deployment test running in Postmates' infrastructure with this PR in place. I'm hoping it'll surface some edge cases. I do want to better understand the errors that the underlying library will produce and bail out for absolute failures, else we'll just loop forever here. Also need to document the module and test the configuration parsing.

build-container.sh
```diff
@@ -0,0 +1,30 @@
...
+function usage() {
+    echo "$0 <version>"
+}
+
+if [ -z "$1" ]; then
+    usage
+    exit 1
+fi
+
+VERSION="${1}"
```
benley: fyi: `VERSION=$1` is good enough; simple assignment isn't subject to globbing or word-splitting.

build-container.sh
```diff
@@ -0,0 +1,30 @@
...
+mkdir -p target/bindir/
+[ -f target/bindir/confd-bin ] || curl -L "$confd_url" > target/bindir/confd-bin
+chmod +x target/bindir/confd-bin
+
+docker build -t cernan-build -f docker/build/Dockerfile .
+CONTAINER_ID=$(docker create cernan-build)
+# Remove 'cernan' from the build image 
+docker container cp ${CONTAINER_ID}:/source/target/release/cernan target/bindir/
+docker rm ${CONTAINER_ID}
+docker build --build-arg EXE_ROOT=target/bindir -t quay.io/postmates/cernan:latest -t quay.io/postmates/cernan:${VERSION} -f docker/release/Dockerfile .
```
benley: you may want to add quotes here, like `-t "quay.io/postmates/cernan:$VERSION"`

Issue #162, opened 2016-12-16
Make CKMS guaranteed error configurable


blt:
Right now every internal CKMS is hard-coded to have an error no greater than 1/1000. We should make this configurable by the user.

blt:
This has been done. Shipped with 0.8.0.

PR #274, opened 2017-06-18
Telemetry guaranteed error configurable

gliush:
It's an attempt to fix https://github.com/postmates/cernan/issues/162
The diff is quite large, though, all the changes to all the files looks very similar. I'd rather checked it by commits.
The most important things to note:
- The change looks very similar to what was needed for 'flush-interval' to work. Though, for the current change we don't need separated configuration for all the sinks/sources, the one global value is used.
- I'm not sure that I've picked the best way to do it

### DIFF
--- a/examples/configs/basic.toml
+++ b/examples/configs/basic.toml
@@ -3,6 +3,8 @@ scripts-directory = "examples/scripts/"
 
 flush-interval = 5
 
+telemetry-error = 0.005
+
 [tags]
 source = "cernan"
 
--- a/examples/configs/basic.toml
+++ b/examples/configs/basic.toml
@@ -3,7 +3,7 @@ scripts-directory = "examples/scripts/"
 
 flush-interval = 5
 
-telemetry-error = 0.005
+telemetry-error-bound = 0.005
 
 [tags]
 source = "cernan"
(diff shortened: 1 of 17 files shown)


gliush:
@blt: Could you have another look?
I'm not sure about the following:
- Current implementation requires `DEFAULT_TELEMETRY_ERROR_BOUND` constant in 3 modules, not very nice.
- Instead, we might have a Option<f64> as a telemetry_error_bound (to be replaced in Telemetry module with the default value), but it's also not very clear.
- I had to add another report_telemetry with 3 arguments and to increase readability I renamed all the report_* functions to have a <NAME><ARITY> pattern: report_telemetry2, report_telemetry3, report_telemetry5
Thanks in advance.

blt:
@gliush will do!

gliush:
Yep, that makes sense. Actually, it's the same as my 2nd question (to have an `Option<f64>` as a 3d parameter to Telemetry::new).
Though, your approach will require less changes to the original sources, so we should prefer it.
Will do.

blt:
Cool! 

Most rust projects I've seen – since Rust lacks default parameters to functions – prefer the use of differently named constructors. It's a little goofy but keeping cernan close to idiomatic seems worth doing to me.

blt:
I'm going to close this out since it's pretty far out of date from current master. Still interested in the changes though!

PR #277, opened 2017-06-21
Address Elasticsearch 0.6.1 issues

### LINKED ISSUE #275: Maintain parity with firehose telemetry for ES
The ElasticSearch sink should provide an equivalent to: `cernan.sinks.firehose.records.total_delivered`

### LINKED ISSUE #276: Elasticsearch does not ship tags
In the firehose sink we inspect the tags available in the logline and ship those with the payload. We do not do this in the elasticsearch sink but must.

blt:
Please see individual commits for details.

Closes #275. Closes #276.


src/lib.rs
```diff
@@ -34,16 +34,14 @@ extern crate rusoto_core;
 extern crate rusoto_firehose;
 extern crate seahash;
 extern crate serde;
+#[macro_use]
```
pulltab: What effect does moving this have?  Just tidy?

Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "MIT"
 name = "cernan"
 readme = "README.md"
 repository = "https://github.com/postmates/cernan"
-version = "0.6.1"
+version = "0.6.2-pre"
```
tsantero: do we plan on keeping this versioned as a pre-release?

src/lib.rs
tsantero: @pulltab looks like a result of removing `elastic_types_derive`, as you'll see later in the code the type of the buffer in ES is now simply a `Vec`

Cargo.toml
blt: I've been marking new PRs with `-pre` if we've just come off a stable release. This just signals that it's not a full release. In the past we had 'stable' releases by version but with a distinct SHA.

tsantero: seems reasonable.

src/lib.rs
blt: @pulltab @tsantero now that we're using the `json!` macro we need to import that macro from `serde_json`. Previously we weren't using it.

tfheen:
Hooray for switching to the bulk API, from my reading of this.

Issue #247, opened 2017-04-11
Wavefront sink doesn't report zero counts, making accurate graphs difficult


bitglue:
Wavefront doesn't interpret missing points as zeros: it interpolates the points. So when the points are something like `1, nil, nil, 1, nil, 1`, Wavefront will interpret that under some contexts as `1, 1, 1, 1, 1, 1` which of course is very wrong.

The `default()` function fixes this somewhat, but only at high zoom levels. `default` will only fill in a point if it's missing in a bucket, but after zooming out such that the Wavefront bin width is wider than the interval between the points, the results become inaccurate.

blt:
I reached out to Wavefront to get their comments on sending zeros. Excerpts follow: 

> Yes having cernan send zeros is one option to help create a consistent graph. However this can be dangerous in times where the metric being reported is a counter and sending zero points will not be a true indication of the level of counting that is occurring.

With regard to cernan's [internal model](https://github.com/postmates/cernan/wiki/Data-Model#aggregation) for ephemeral SUM | SUMMARIZE this is fine. A SET probably ought to still kick a nil toward wavefront. This isn't an issue for persistent telemetry. The [Sources](https://github.com/postmates/cernan/wiki/Data-Model#aggregation) page discuss how different ingest protocols map to the internal model. 

> As for timeseries not being consistent across all zoom levels. The issue is with the bucketing of points. In the chart menu you can select what kind of aggregations are used in buckets. A bucket is basically the aggregation of several points into one point when for example the zoom level is very large (over a week) and the source is reporting every second. There is no way that you could see points representing 1-second on your computer monitor when there is 600k seconds in a week. So we bucket points depending on your computer resolution and browser window size to help display the data. I cannot make a general recommendation on how to have more consistent values across all zoom levels until I see an example.

> A general rule of thumb though is to change the aggregation under Chart > General > Summarize By setting to fit the kind of data you want to see at all levels of zoom. Max and Min are quite consistent across time windows for time series with large spontaneous spikes or deep troughs as well as counters. Average can be quite consistent for timeseries that oscillate around a specific value.

These comments make me wonder how effective filling in the blanks will be. Easy enough to make enabling zeros a configurable thing.

bitglue:
That doesn't make any sense. Maybe they thought "counter" means the sort that is increasing, where `rate()` would be appropriate?

blt:
I'm honestly not sure. I'll see about getting that clarified.

blt:
[...]

PR #283, opened 2017-07-15
Wavefront zero padding

blt:
This pull-request introduces 'padding' into the wavefront sink. The behaviour is discussed [here](https://github.com/postmates/cernan/issues/247#issuecomment-314837565). The main show is in the property tests for the new padding iterator, as well as the commented iterator routine itself. 

Rust iterators are an odd breed. The padding iterator, specifically, is an "into" iterator, meaning that it takes ownership of the original iterator's members. You'll note that this means we have to be careful not to drop the formatted wavefront payload if there's a flush error because it cannot be recreated once that payload goes out of scope. 

This PR does _not_ address padding across flushes. Consider in the original issue a sequence like so was proposed: `[1,1,nul,1] -> [1,1,0,nul,0,1]`. If there were a flush after the nul `[1,1,nul,FLUSH,1]` then one interpretation of #247 would be to emit `[1,1,0,nul,0,1]` still. This implies added storage so that the sink can 'remember' what it's flushed previously. Also, when do we properly consider a sequence to terminate? That is, if a sequence is not reported on for N flushes is there some value of N for which that missing, departed sequence should be laid to rest? At present N = 0. 

This PR does _not_ address eliding extra zeros. That will come in a later pull request.

Please see individual commits for details. 

Part of the work to address #247

### DIFF
--- a/src/sink/wavefront.rs
+++ b/src/sink/wavefront.rs
@@ -93,7 +93,6 @@ pub fn padding<I: Iterator<Item = Telemetry>>(xs: I, span: i64) -> Padding<I> {
     Padding {
         span: span,
         orig: xs,
-        last_emit_ts: None,
         emit_q: Vec::new(),
     }
 }
@@ -101,7 +100,6 @@ pub fn padding<I: Iterator<Item = Telemetry>>(xs: I, span: i64) -> Padding<I> {
 pub struct Padding<I> {
     span: i64,
     orig: I,
-    last_emit_ts: Option<i64>,
     emit_q: Vec<Telemetry>,
 }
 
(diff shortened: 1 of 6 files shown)


blt:
@lynchc relevant to your interests.

src/sink/wavefront.rs
```diff
@@ -392,7 +570,7 @@ mod test {
         wavefront.format_stats(dt_2);
         let lines: Vec<&str> = wavefront.stats.lines().collect();
 
-        println!("{:?}", lines);
+        //println!("{:?}", lines);
```
tsantero: remove instead of comment out?

src/sink/wavefront.rs
```diff
@@ -285,13 +378,98 @@ impl Sink for Wavefront {
...
+                    // three possibilities:
+                    //
+                    //  1. the points don't hash the same, so we move on
+                    //  2. the points do hash the same:
+                    //     a. if their timestamps are greater than one span
+                    //        apart then they are both zero
+                    //     b. if both points are non-zero they must not be
+                    //        more than one span apart
+                    if t.hash() == next_t.hash() {
+                        let span = (t.timestamp - next_t.timestamp).abs() / (bin_width as i64);
```
tsantero: remind me, how does rust handle rounding conversion with `as i64` ?

src/sink/wavefront.rs
```diff
@@ -215,6 +308,7 @@ impl Sink for Wavefront {
     }
 
     fn flush(&mut self) {
+        self.format_stats(time::now());
```
tsantero: since chrono uses `Duration`, shouldn't this be `time::Instant::now()`?

src/sink/wavefront.rs
blt: Accidental comment. This is useful in test as the lines are only printed out on failure. Will correct.

src/sink/wavefront.rs
blt: In this sink we don't actually use the passed in time. I'll remove it altogether.

tsantero: ah, you're right.

blt: I believe you're correct, that said, it should be `Instant::now` to have the same monotonicity.

src/sink/wavefront.rs
blt: In this case bin_width :: u8 so rust'll just pad the underlying bytes to hit i64. I'm sure somewhere in [The Reference](https://doc.rust-lang.org/reference/) there's info on exactly how this works but since they split it up into chapters it's become impossible to search. :/

blt: Ah! Found it in the nomicon, [here](https://doc.rust-lang.org/nomicon/casts.html).

> For numeric casts, there are quite a few cases to consider:
> 
> * casting between two integers of the same size (e.g. i32 -> u32) is a no-op
> * casting from a larger integer to a smaller integer (e.g. u32 -> u8) will truncate
> * casting from a smaller integer to a larger integer (e.g. u8 -> u32) will
> * zero-extend if the source is unsigned
> * sign-extend if the source is signed
> * casting from a float to an integer will round the float towards zero
>     * NOTE: currently this will cause Undefined Behavior if the rounded value cannot be represented by the target integer type. This includes Inf and NaN. This is a bug and will be fixed.
> * casting from an integer to float will produce the floating point representation of the integer, rounded if necessary (rounding strategy unspecified)
> * casting from an f32 to an f64 is perfect and lossless
> * casting from an f64 to an f32 will produce the closest possible value (rounding strategy unspecified)
>     * NOTE: currently this will cause Undefined Behavior if the value is finite but larger or smaller than the largest or smallest finite value representable by f32. This is a bug and will be fixed.

tsantero:
@bitglue I'm satisfied with this changeset. Would you mind adding a review so that @blt can progress forward with merging in the `0`s in wavefront flushes?

blt:
Per conversation with @bitglue I'm going to chase this PR with another one which will add 'sustain' configuration to the wavefront sink. For N flush intervals after the last point we'll emit a zero pad on either end of a window for a time series which doesn't have a point in that window. By default N will be zero.

PR #284, opened 2017-07-17
Keep sockets in native / wavefront

blt:
The previous implementation of these sinks created and destroyed sockets
on each flush. There is no good reason for this any longer. We now take
the approach of preserving the socket only so long as we're able to send
information through it, otherwise we destroy the socket and get a new one.

### DIFF
--- a/src/sink/wavefront.rs
+++ b/src/sink/wavefront.rs
@@ -20,6 +20,7 @@ pub struct Wavefront {
     percentiles: Vec<(String, f64)>,
     pub stats: String,
     flush_interval: u64,
+    stream: Option<TcpStream>,
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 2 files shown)


blt:
@bitglue relevant to your interests

bitglue:
Is this what's causing the RST every 10 seconds? Or was that the firehose sink?

blt:
@bitglue I _believe_ it's the RST sink, but we don't control that socket. It's hard to fix. On inspection of the codebase I realized we could fix this too.

bitglue:
By "RST sink" you mean "Firehose sink"?

blt:
@bitglue I meant firehose sink, yes. I have _no idea_ why I wrote "RST sink". That is gibberish.

src/sink/wavefront.rs
```diff
@@ -317,43 +352,24 @@ impl Sink for Wavefront {
...
+                    self.aggrs.reset();
+                    self.stats.clear();
+                    self.delivery_attempts = 0;
+                    return;
+                } else {
+                    self.delivery_attempts = self.delivery_attempts.saturating_add(1);
+                    delivery_failure = true;
                 }
+            } else {
+                time::delay(self.delivery_attempts);
```
tsantero: I can't tell where `time::delay` is coming from. Is the fundamental unit passed in seconds? Perhaps maybe we want `std::thread::sleep` instead here?

Also, it seems like it's possible for this to be hit when `self.delivery_attempts` is `0`, which I don't think is a behavior we want?

The backoff logic here is linear. Would we prefer exponential?

blt: `time::delay` comes from [here](https://github.com/postmates/cernan/blob/master/src/time.rs#L38). The inputs are linear but the backoff occurs as the power of two, up to a cap. We special-case the `0` to be a no-op https://github.com/postmates/cernan/blob/master/src/time.rs#L40.

tsantero: this all makes so much more sense when you actually read the function being called 😂

PR #288, opened 2017-07-19
Fix wavefront persist

### LINKED ISSUE #287: Persist Telemetry is broken in the Wavefront sink
With the switch to padding via consuming aggregator the Wavefront sink has broken its persist metrics. These are no longer persisted across flush windows. This must be corrected.

blt:
This commit ensures that we keep our persist telemetry around
for the next flush window. Much thanks to Tom Santero for the
simple fix idea.

This commit fixes #287

### DIFF
--- a/src/sink/wavefront.rs
+++ b/src/sink/wavefront.rs
@@ -146,11 +146,15 @@ where
                             if x.value() == Some(0.0) {
                                 self.emit_q.push(y);
                             } else {
-                                let sub_y = y.clone().timestamp(y.timestamp - 1).set_value(0.0);
+                                let sub_y = y.clone()
+                                    .timestamp(y.timestamp - 1)
+                                    .set_value(0.0);
                                 self.emit_q.push(y);
                                 self.emit_q.push(sub_y);
                                 self.emit_q.push(
-                                    x.clone().timestamp(x.timestamp + 1).set_value(0.0),
+                                    x.clone()
+                                        .timestamp(x.timestamp + 1)
+                                        .set_value(0.0),
                                 );
                             }
                             return Some(x);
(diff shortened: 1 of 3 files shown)


src/sink/wavefront.rs
```diff
@@ -247,7 +251,13 @@ impl Wavefront {
 
         let mut tag_buf = String::with_capacity(1_024);
         let aggrs = mem::replace(&mut self.aggrs, Buckets::default());
-        for value in padding(aggrs.into_iter().filter(|x| x.is_zeroed()), self.bin_width) {
+        for value in
+            padding(aggrs.into_iter().filter(|x| x.is_zeroed()), self.bin_width)
```
pulltab: Can you unpack this a bit for me?

blt: Sure thing. 

`aggrs.into_iter()` converts `aggrs` into a iterator, meaning `aggrs` is no longer a valid reference and we'll be given an iterator that owns the `Telemetry` formerly in `aggrs`. 

`.filter(|x| x.is_zeroed()` removes any zero valued `Telemetry` from the iterator stream. We do this in preparation for padding. 

`padding` is a function which returns a `Padding` iterator, one that adds zeros according to the procedure outlined here: https://github.com/postmates/cernan/issues/247#issuecomment-314837565 We previously removed all zero-values from the stream to avoid complicating the padding iterator. It's guaranteed that there are no zeros to have to cope with.

src/sink/wavefront.rs
```diff
@@ -247,7 +251,13 @@ impl Wavefront {
...
+        {
+            if value.persist {
+                let new_val = value.clone();
+                self.aggrs.add(new_val.timestamp(value.timestamp + 1));
```
pulltab: Oh..this is self.aggrs.

blt: It is. `self.aggrs` is different from `aggrs` which got consumed previously. When we do `mem::replace` above what we're doing is allocating a new `Bucket`, sticking it into the reference at `self.aggrs` and returning `aggrs`. 

That is, `self.aggrs` is a pointer into the `Self` == `Wavefront` struct and `aggrs` is a locally scoped heap object.

PR #292, opened 2017-07-31
Apply clippy suggestions, Cargo.lock updates

blt:
This is a breeze-through of the codebase to apply clippy suggestions.
There's no real code change here.


tsantero:
did you forget to push up `Cargo.toml`

blt:
@tsantero `Cargo.toml` didn't change, I ran `cargo update`. 

That is kind of confusing and _is_ maybe a problem. We pin cernan deps to within minor point changes – I don't recall any accidental breakage – but it's not impossible either. We're trusting everyone downstream.

tsantero:
@blt ah, makes sense. Your commit message mentioned `Cargo.toml` and I didn't bother looking at the diff on the lock file. Just wanted to be sure

blt:
@tsantero ooooh I'm a dum dum. I meant `Cargo.lock` but said `.toml`.

PR #293, opened 2017-08-04
Integrate histogram

blt:
This is a slab of work to get quantiles/Histogram integrated into cernan. There's in-progress work on quantiles as well, as you can see from the TOML. The work has been slow-going so I'm putting it up here now for preliminary review. You'll find `unimplemented!()` and other bad bits. Please _do_ flag those. We don't want to forget them.


blt:
@tsantero @pulltab as of ee4819d I've realized we've broken the zero padding in the wavefront sink with these changes.

blt:
@tsantero @pulltab No, unrelated. I believe the defect was present prior to this PR and should be resolved after its merged.

Issue #295, opened 2017-08-12
Add `delay_tolerance`, `flush_boundary_tolerance` concepts


blt:
The work on #293 has surfaced an old issue that we punted on. Consider this sequence of events:

```
T0 
FLUSH
T0
T1
```

where `T` is some telemetry at some time. What you'll see in outputs is two distinct `T0` points, because they fall on either side of the flush. What you would _like_ to see in some occasions is a single `T0` aggregated from the two inputs. 

Here's what I propose to do. We need two things in cernan to make this possible. First, we must have a `flush_boundary_tolerance`, that is, a mechanism by which we keep back points – even if a Flush has come through – until we're reasonably sure that no more points will arrive for that time. Second, we need a `delay_tolerance`, a mechanism by which we reject points that are too old. If these mechanisms are configurable the end-user can decide what their tolerances are for these two setups. 

We should add two new filters to cernan – one for each of the above mechanisms. That way we don't have to complicate existing sinks and users don't pay the cost for what they don't use.

blt:
Closed by #297

PR #297, opened 2017-08-12
Introduce flush_boundary_filter

blt:
The idea with this filter is that we want to be able to hold back
points for some specified tolerance before they're sent to flush.
This resolves the split time series problem discussed in #295


blt:
This closes #293.

PR #298, opened 2017-08-16
Add Documentation, Moves to Quiet Clippy 

blt:
This pull-request is intended to quiet clippy by taking its suggestions. We've also flipped the no-docs lint to on as we've now worked things for that lint to be quiet. A great many lines have changed but it's all semi-mechanical changes.


blt:
Self-merging because this'll surely stink to review.

Issue #128, opened 2016-10-13
Overhaul cernan's reliance on small allocations


blt:
We've gotten lazy in cernan for expediency purposes. While this has been okay for the time being we're cloning like crazy. What we need are explicit lifetimes in places to begin using Cow where sensible. 

Unfortunately, the lifetimes that cernan gets automatically are, apparently, pretty complicated and I'm not the best at their algebra. There's no way to get Rust to print the lifetimes it generates so... uh, bit of an R&D task.

blt:
In the meantime [string-cache](https://github.com/servo/string-cache/) has been coming along. We are constrained to meet serialization as imposed by [hopper](http://github.com/postmates/hopper), in the event we get overburdened and have to start queuing to disk. As of this writing string-cache is at serde 0.9 but once it's on 1.0 we ought to be in a better place. 

We'll have to use a dynamic cache and there's some possibility of denying service by flooding cernan with nonsense strings but I _guess_ that's okay? Need to balance the untrusted vs. trusted assumptions somehow.

blt:
This will always be an ongoing project. Closing out since there's no end game per se to be had.

PR #299, opened 2017-08-17
Introduce String Cache

blt:
This pull-request is something we've been working toward for a while. #128 is the primary tracking issue. What we've done is introduce a straightforward string cache and have removed the small allocations in `Telemetry` and `LogLine` by forcing the name, path, tagmap to keep references to the cache, rather than the alloc'ed string itself. 

We don't do any fancy automatic dereferencing and the extra locking needed to make this work needs to be stress tested under realistic conditions.


blt:
Experimentation demonstrates that this has worse performance than just using jemalloc. Oh well.

PR #302, opened 2017-09-01
Padding control for wavefront sink

### LINKED ISSUE #301: Don't pad SET in wavefront sink
As of 0.7.0 we've got padding introduced into the wavefront sink. The feedback we've gotten is that for SET – gauge, ephemeral or not – is that padding makes things weird in Wavefront. 

Introduce a config flag to avoid padding.

blt:
This commit adds configuration knobs in the wavefront sink to turn zero-padding on and off by aggregation type. Padding was introduced in 0.7.0 but there's been concern with the results now that 0.7.1 is out and about. We now allow users to decide to pad or not.

By default this feature is OFF for all aggregations. Research by @pulltab suggests that neither etsy/statsd nor statsite have this feature.

This resolves #301

### DIFF
--- a/src/sink/wavefront.rs
+++ b/src/sink/wavefront.rs
@@ -34,6 +34,23 @@ lazy_static! {
     pub static ref WAVEFRONT_VALVE_OPEN: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
 }
 
+#[derive(Copy, Clone, Debug, Default, Serialize, Deserialize)]
+/// Controls which aggregrations will be padded
+pub struct PadControl {
+    /// DO pad SET Telemetry if the value is true, DO NOT pad if the value is
+    /// false.
+    pub set: bool,
+    /// DO pad SUM Telemetry if the value is true, DO NOT pad if the value is
+    /// false.
+    pub sum: bool,
+    /// DO pad SUMMARIZE Telemetry if the value is true, DO NOT pad if the value is
+    /// false.
+    pub summarize: bool,
+    /// DO pad HISTOGRAM Telemetry if the value is true, DO NOT pad if the value is
+    /// false.
+    pub histogram: bool,
+}
+
 /// The `wavefront` sink emits into [Wavefront](http://wavefront.com), a
 /// proprietary metric aggregation and alerting product.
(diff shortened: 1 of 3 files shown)


src/sink/wavefront.rs
```diff
@@ -34,6 +34,23 @@ lazy_static! {
     pub static ref WAVEFRONT_VALVE_OPEN: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
 }
 
+#[derive(Copy, Clone, Debug, Default, Serialize, Deserialize)]
+/// Controls which aggregrations will be padded
+pub struct PadControl {
+    /// DO pad SET Telemetry if the value is true, DO NOT pad if the value is
```
pulltab: As padding is seemingly unique to cernan can you put in an explanation about what we mean by padding?

Perhaps in the documation for the wavefront sink?

src/sink/wavefront.rs
```diff
@@ -151,10 +173,15 @@ impl<'a> Pad<'a> {
         }
     }
 
-    pub fn is_pad_zero(&self) -> bool {
+    pub fn skip_pad(&self, pad_control: &PadControl) -> bool {
         match *self {
             Pad::Zero(_, _) => true,
-            Pad::Telem(_) => false,
+            Pad::Telem(x) => match x.kind() {
+                AggregationMethod::Histogram => !pad_control.histogram,
```
pulltab: Why are these being negated?

src/sink/wavefront.rs
blt: Good idea. I'll update the wiki with the configuration option, padding discussion once this PR is merged. Work?

src/sink/wavefront.rs
blt: The PadControl structure is affirmative, if the value is true then we pad. Here we're deciding whether to skip adding a pad, so we need the opposite truth value. That make sense?

pulltab: Got it.

src/sink/wavefront.rs
pulltab: Good for me.

PR #303, opened 2017-09-01
Introduce reading tags from environment variables

### LINKED ISSUE #279: Make default tags a thing
It would be convenient for some users if cernan had a default set of tags. @blakebarnett is the primary contact for this feature. For users that do not use tags we'd want to add a `default-tags: Bool` option to cernan to avoid tags, false by default. 

Default tags we have so far:

* hostname 

Open to suggestions. 

- - - 

After further discussion, we now want default-tags to be a list, not a boolean. See https://github.com/postmates/cernan/issues/279#issuecomment-312074665 for more details.

blt:
This commit introduces the ability to configure a tag from an
environment variable. This was requested by @blakebarnett. It
should make running cernan in kubernetes nicer.

Resolves #279

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -7,6 +7,7 @@ use clap::{App, Arg};
 use metric::TagMap;
 use rusoto_core::Region;
 use std::collections::HashMap;
+use std::env;
 use std::fs::File;
 use std::io::Read;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


blt:
@tfheen @blakebarnett relevant to your interests.

blt:
@tfheen 

> The change itself looks good to me, but maybe also update an example or some docs?

Will do. I'll update the wiki once this change is merged into master. Every so often I think about including cernan's documentation in source control but haven't done so. Not entirely sure what generator I'd go with, anyway.

> I do wonder if this is what would be best, though. What about if it just iterated through CERNAN_TAGS_* and used that (or a space-separated CERNAN_TAGS in the environment) without having to specify it in the configuration file?

I think the expectation was that cernan should be able to pick up existing environmental variables. This approach -- which I'd be happy to get a patch for, FYI -- would require duplication of existing variables into the CERNAN_TAGS_* space.

tfheen:
>  Not entirely sure what generator I'd go with, anyway.

I'd go with "pick one".  Picking one and having the docs in the source is (IMO) much more valuable than exactly which one you pick. If you get it _terribly_ wrong you can always change it later, but most likely it'll be at least ok.

>  which I'd be happy to get a patch for, FYI -- would require duplication of existing variables into the CERNAN_TAGS_* space.

I think maybe just `CERNAN_TAGS="foo bar baz"` would be fine, or do we need to support tags with spaces in their name? I suspect that'll break in hilarious ways pretty quickly once they hit other systems.

blakebarnett:
I'm happy to have a list of tags like this also, I can make k8s provide the $HOSTNAME as a part of the list too, but IMO it's less obvious that way, but certainly less configuration heavy.

blt:
I think I'll keep on with the existing implementation but I'm very happy to receive patches for alternative approaches.

PR #305, opened 2017-09-05
Drain unbounded buffer in Internal source

### LINKED ISSUE #260: Internal source has a potentially unbounded buffer
The internal source, if not given any forwards, will store its members indefinitely. We need to make it so that in the event of there being no forwards for this source it simply drops elements, instead of keeping them in the `VecDeque`.

blt:
This commit drains the unbounded buffer in the Internal source in
the case where there's no channels to push Internal metrics into.
This resolves #260

### DIFF
--- a/src/source/internal.rs
+++ b/src/source/internal.rs
@@ -316,10 +316,10 @@ impl Source for Internal {
                     }
                 }
             } else {
-                // do nothing, intentionally
+                while let Some(_) = Q.pop() {
+                    // do nothing, intentionally
+                }
             }
-
-
         }
     }
 }
--- a/src/source/internal.rs
+++ b/src/source/internal.rs
@@ -316,6 +316,9 @@ impl Source for Internal {
                     }
                 }
             } else {
+                // There are no chans available. In this case we want to drain
+                // and deallocate any internal telemetry that may have made it
+                // to Q.
                 while let Some(_) = Q.pop() {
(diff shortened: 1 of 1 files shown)


src/source/internal.rs
```diff
@@ -316,10 +316,10 @@ impl Source for Internal {
                     }
                 }
             } else {
```
pulltab: Can you add a comment here?  The branch above is so large it isn't immediately obvious to which cases this else applies.

blt: Good idea.

src/source/internal.rs
```diff
@@ -316,10 +316,10 @@ impl Source for Internal {
...
-                // do nothing, intentionally
+                while let Some(_) = Q.pop() {
```
pulltab: Is there a chance we block here?

blt: There isn't. The internal implementation of `coco::Stack` might take more or less time to return an item from the stack but when an item is not available `None` will be returned and we'll exit the loop. 

Now, that said, there is the possibility that either of these loops -- in this branch or the one, similar, above -- will be saturated by points and will be unable to exit. I was unable to trigger this condition even with a custom-built source whose sole role was spamming the Internal stack. I believe this is a low-probability condition and, even so, is something that's inline with "process all the points available".

Issue #290, opened 2017-07-21
Cernan needs a proper binning histogram


blt:
As of 2f08dbc0c11f0089e6e798241045a76730c3559a the only summarization type that cernan supports is the [CKMS](https://github.com/postmates/quantiles#ckms---effective-computation-of-biased-quantiles-over-data-streams) from our quantiles. Quantile structures are great but suffer from the need to aggregate the structures before their data makes sense, as in Wavefront or Prometheus. This is _not_ a problem for binning histograms, that is histograms computed by using a client-supplied min/max range plus number of bins. Individual cernan instances are able to choose the same aggregating behaviour without coordination. 

We should extend the native protocol to include a HISTOGRAM summary type. I believe we will need: 

* a global configuration for default min/max
* a global configuration for default number of bins
* extension to the native protocol to include stream-specific min/max, bins

We may also choose to retarget statsd histogram/timers to this type. I'm unsure about this.

This is related to #66. 

@lynchc @bitglue relevant to your interests

bitglue:
If we're re-targeting statsd, I'd be more inclined to do that at the client level to have the existing application code send histograms for the `.timer()` function, than change the statsd protocol semantics.

blt:
@pulltab ^^

blt:
@bitglue one thing I'm not clear about, if we have application code send histograms (as multiple counters?) is how we then interact with prometheus, since it does have a native histogram type. I guess we can send Prometheus multiple counters too but if we can get clients to send us that extra info I don't see where we come out behind.

tfheen:
The individual bin width should be configurable. It's interesting to see if something takes 5 or 10ms, but probably not that interesting to know if it takes 1005 or 1010 ms.

blt:
@tfheen looking at just the native protocol, my expectation is that the user will send the binning information along with each payload. Under discussion now is whether we extend that same behaviour into the statsd source by accepting a broader form of statsd payloads.

bitglue:
The Prometheus data model for histograms is also multiple counters. There's just a bit of metadata that specifies the bin boundary. Prometheus uses the metadata for a function which estimates quanties based on the histogram. Prometheus clients calculate the binning and expose multiple counters with metadata.

Example: https://github.com/prometheus/client_python/blob/master/prometheus_client/core.py#L862-L868

blt:
This is now complete and is documented.

Issue #285, opened 2017-07-18
Prometheus time window scraping support


lynchc:
Rather than having cernan dump all metrics collected in memory on a prometheus scrape, I would like it to present collected metrics for a configurable amount of time. 

The goal here is so that we can point multiple prometheus scrapers at the cernan endpoint and duplicate the results for HA

blt:
@lynchc my thinking is that I'll add a moving window of configurable size to the underlying storage. Metrics will only drop off when they've gotten too old, not when they've been read. @benley does that work for you too?

benley:
That sounds OK, as long as the metrics are all things that cease to be meaningful if they get too old.  If cernan is acting as an event receiver for things that happen like once an hour and get exported as counters or gauges, you probably wouldn't want those to expire.

blt:
Those would end up expiring, yeah. Say the moving window is 60s and the event is [graphite telemetry](https://github.com/postmates/cernan/wiki/Sources#graphite). Cernan will store that as a an [ephemeral, SET](https://github.com/postmates/cernan/wiki/Data-Model#aggregation) and drop it after 60s of receipt. A [statsd gauge](https://github.com/postmates/cernan/wiki/Sources#statsd) would persist until it gets reset by some other value. 

I guess 'too old' is really a function of input protocol and persistence flags set via that protocol.

benley:
@brian-brazil suggested that we take a look at how the Prometheus [statsd exporter](https://github.com/prometheus/statsd_exporter) and [graphite exporter](https://github.com/prometheus/graphite_exporter) handle this, because there is a fair bit of functional overlap with cernan there.

blt:
@benley @brian-brazil got it. I don't read Go all that well so correct me if I'm wrong here. 

It looks like what the graphite exporter does is store up samples, notes the timestamp and for some tick interval and expiration age-limit it goes through all the samples and removes the toasted ones: 

  *  https://github.com/prometheus/graphite_exporter/blob/master/main.go#L145 

I can't figure out the structure of the statsd example. I could use some help there. Near as I can tell it's creating prometheus values directly on receipt? 

  * https://github.com/prometheus/statsd_exporter/blob/master/exporter.go#L76

For the prometheus sink it's storing samples 

  * https://github.com/prometheus/graphite_exporter/blob/master/main.go#L171 

[...]

PR #306, opened 2017-09-07
Make Prometheus sink obey aggregations

blt:
This commit makes the prometheus sink send a variety of types, rather
than the quantile, now that Telemetry is backed by different and
distinct kinds.

This is part of the work for #290.

This is related to the work of #285.


blt:
@lynchc https://github.com/postmates/cernan/pull/306/commits/d39fba1e40bcd09312f69ad3e6166e4bb3aaf6b0 adds documentation which, I think, addresses some of your questions. Let me know if you have more?

src/sink/prometheus.rs
```diff
@@ -92,40 +117,52 @@ struct PrometheusAggr {
...
-) -> Option<::std::cmp::Ordering> {
-    match l.name.partial_cmp(&r.name) {
-        Some(::std::cmp::Ordering::Equal) => ::metric::tagmap::cmp(&l.tags, &r.tags),
-        other => other,
-    }
+    // need to meet it there. But! We can't just drop metrics as soon as they're
+    // read, apparently, since prometheus expects to be able to re-read multiple
+    // times and get roughly similar results. What we do, then, is store a
+    // sliding window of metrics per hash, keeping only ten values per
+    // timestamp. We sum these values up at report time.
```
benley: Is `ten values per timestamp` meant to be a reference to `retain_limit`?

benley:
This is only tangentially related to this PR, but would it make any sense to have cernan use something like https://github.com/pingcap/rust-prometheus in the prometheus sink?

src/sink/prometheus.rs
```diff
@@ -1,11 +1,30 @@
...
+//! following method:
+//!
+//!   - SET -> gauge
+//!   - SUM -> counter
+//!   - QUANTILES -> summary 
+//!   - HISTOGRAM -> histogram
+//!
+//! All points are retained in a sliding window defined by
+//! `retain_limit`. Points will persist indefinately unless new points come
+//! behind them to push them out of the `retain_limit` window.
```
benley: Does `retain_limit` aggregation apply to `gauge` metrics as well?  If so, how does that work?

blt:
> This is only tangentially related to this PR, but would it make any sense to have cernan use something like https://github.com/pingcap/rust-prometheus in the prometheus sink?

@benley I don't believe so. We'd have to knock it into a different shape to fit the existing Sink trait, plus its inclusion only affects the aggregation implementation of the sink, which requires cernan to map from its internal model to another, distinct model.

src/sink/prometheus.rs
blt: It applies to all cernan aggregation methods, yes. If a prometheus aggregation should be treated differently do let me know. Otherwise, points persist indefinitely until a new point comes along to knock it out of place.

src/sink/prometheus.rs
blt: Yes. That should be corrected.

src/sink/prometheus.rs
benley: As long as cernan is giving prometheus the most recent value of the gauge and not doing something weird like summing or averaging the last 10 points, it sounds harmless :-)

blt: What happens is we sum all of the values that we're storing: https://github.com/postmates/cernan/pull/306/files#diff-497dac92ece3082c6719be9ac7ca33e2R138

AddAssign for Telemetry is defined here: https://github.com/postmates/cernan/blob/da80ac412cc2bfccbea6b157c67f07f3744089a7/src/metric/telemetry.rs#L213

Add for Value is defined here: https://github.com/postmates/cernan/blob/da80ac412cc2bfccbea6b157c67f07f3744089a7/src/metric/telemetry.rs#L79

blt: The specific line you want is here: https://github.com/postmates/cernan/blob/da80ac412cc2bfccbea6b157c67f07f3744089a7/src/metric/telemetry.rs#L95

LHS (SET | SUM) + RHS (SET) = RHS (SET)

benley: Oh, I think I see what was confusing me.  This is specifically how to handle _aggregating_ multiple sources of a single gauge metric, right?  If so, the ideal case for prometheus would be not to aggregate them at all, but instead to export each value as a separate metric, differentiated by a label (probably `source`).  Then Prometheus can aggregate them according to its configured rules.  Hopefully I'm not totally lost here...

blt: I think we need to make a careful distinction between prometheus' aggregation and cernan Telemetry and how summation over Telemetry works. Points:

1. Prometheus aggregations do not exist until we hit `write_text` or `write_binary` at which point we convert from cernan's Telemetry into the appropriate serialized version. 
2. Cernan aggregations are SET, SUM, HISTOGRAM and SUMMARIZE, defined here: https://github.com/postmates/cernan/wiki/Data-Model#aggregation (HISTOGRAM is left off the documented list until https://github.com/postmates/cernan/issues/290 lands)

The comment above indicates how cernan aggregation methods sum up. In the prometheus sink we keep up to `retain_limit` time bins worth of Telemetry and then sum over them to get the Telemetry we'll report out to prometheus. A SUM will be SUM(x_0 + x_1 + ... x_retain_limit) for instance. A SET will be SET(x_retain_limit). 

All that said, I'm not sure what you mean to suggest by 

> If so, the ideal case for prometheus would be not to aggregate them at all, but instead to export each value as a separate metric, differentiated by a label (probably `source`).

What would the value of `source` be? Some configurations of cernan have a `source` defined in their tag map but not all so I don't imagine that's what you mean.

benley: > What would the value of source be? Some configurations of cernan have a source defined in their tag map but not all so I don't imagine that's what you mean.

`source` would be the name of the instance that sent that value to cernan, presumably.  Hostname of the source IP, unless there is a better name to use somewhere.

blt: > `source` would be the name of the instance that sent that value to cernan, presumably. Hostname of the source IP, unless there is a better name to use somewhere.

Got it. These things are only available by configuration, per https://github.com/postmates/cernan/wiki/Configuration#global-tags. Most users do set tags, but not all. 

This doesn't address what cernan should do in the case of a single cernan instance receiving metrics from a single source. Currently, Telemetry is kept according to timestamp -- update to `retain_limit` values -- and then aggregated as discussed.

benley: Actually I might be on the wrong track with the labeling part, at least as a default.  The prometheus statsd exporter doesn't do that, for instance.

However, the part where counters are a sum of only the most recent N values seems extremely wrong to me.  Here's a side by side comparison of sending some counter data to both cernan and an instance of statsd_exporter:

```sh
#!/usr/bin/env bash
# Send some counter values to cernan and statsd_exporter and compare the result

metric_name="test.$$.bar"
values=(5 5 5 5 5 5 5 5 5 5 2 2 2 2 2 5 5 2 5 2 5 5 5 2 2 2 2 1)

echo "Sending ${#values[@]} values:"
echo -n "$metric_name "

for n in "${values[@]}"; do
  echo "$metric_name:$n|c"|nc -w1 -u 127.0.0.1 8125
  echo "$metric_name:$n|c"|nc -w1 -u 127.0.0.1 9125
  echo -n "$n "
done
echo

echo "Cernan result: "
curl -s http://127.0.0.1:8080/metrics|grep "$metric_name"
echo

echo "statsd_exporter result: "
curl -s http://127.0.0.1:9102/metrics|grep "$metric_name"
echo
```

After running the above:
```
Sending 28 values:
test.21981.bar 5 5 5 5 5 5 5 5 5 5 2 2 2 2 2 5 5 2 5 2 5 5 5 2 2 2 2 1 
Cernan result: 
# TYPE test_21981_bar counter
test_21981_bar 31

statsd_exporter result: 
# HELP test_21981_bar_counter Metric autogenerated by statsd_exporter.
# TYPE test_21981_bar_counter counter
test_21981_bar_counter 103
```

Assuming that the metric in question represents something like a total http requests counter, and ignoring the differing metric name (I don't know why stastd_exporter adds _counter when the data type is already explicitly a counter), how does cernan's final value of 31 make any sense?

blt: > how does cernan's final value of 31 make any sense?

That's what this review is for. :) Once upon a time there was an ask to do sliding window – https://github.com/postmates/cernan/issues/285 – but that conversation dead-ended. Now we can get a fix on the semantics. 

Okay, so it looks like, so far:

 - SET should always emit the last value in, forever 
 - SUM should always emit the running sum of values in, forever 

How about HISTOGRAM and SUMMARIZE?

benley: I'm writing up some test cases for these to figure out what to do, sorry for dropping the ball on you again.

blt: @benley cool, no worries. Let me know if I can help.

benley: When combined with the new histogram feature in #320 the histogram and summary features we'll be looking for seem to be in good shape.

blt: @benley Okay, to confirm:

* SET should always emit the last value in, forever 
* SUM should always emit the running sum of values in, forever 
* SUMMARIZE should emit a moving window quantiles summary, `retain_limit` total seconds across
* HISTOGRAM should emit a moving window histogram summary, `retain_limit` seconds across

benley: hmm, I'm still not so sure about the retain_limit part.  In prometheus land, the counters used in histograms and summaries would get reset to 0 when the process exporting metrics restarts.  Cernan doesn't restart when the metric sources restart, but we're also not relying on those counter resets to spot downtime or app restarts.  I think it will make more sense if HISTOGRAMs emit values since forever, like SET and SUM, as the data remains coherent that way. [The prometheus docs](https://prometheus.io/docs/concepts/metric_types/#histogram) seem to concur.

Summaries are trickier, since quantile estimation necessarily involves sliding time windows.  There isn't going a universally applicable decay time, so unless there is a way to configure the retention limit, an arbitrary value will have to do.  Is there any way to make it be a window of N minutes rather than the N most recent values?

blt: > Cernan doesn't restart when the metric sources restart

That'd be a neat trick, but for something like statsd source we'd have to fake it with a liveness timer. 

> I think it will make more sense if HISTOGRAMs emit values since forever, like SET and SUM, as the data remains coherent that way. The prometheus docs seem to concur.

Easily done. 

> There isn't going a universally applicable decay time, so unless there is a way to configure the retention limit, 

What would you be looking for? Configuration per time series? 

> an arbitrary value will have to do. Is there any way to make it be a window of N minutes rather than the N most recent values?

Sure. Every flush pulse we can inspect the summaries and slide the underlying window or windows, if we want to make them configurable per series in some fashion.

benley: Configuration per time series would be cool but I'd settle for a time-based window for the quantiles instead of a count-based one.

One thing I haven't checked but should have: are the `<summaryname>_sum`  and `<summaryname>_count` metrics totals since forever?  If not, can you make them that away?

blt: > Configuration per time series would be cool but I'd settle for a time-based window for the quantiles instead of a count-based one.

I can make that happen, sure. 

> One thing I haven't checked but should have: are the `<summaryname>_sum` and `<summaryname>_count` metrics totals since forever? If not, can you make them that away?

Excellent question. So for SUMMARIZE we're setting the sample sum using value.sample_sum (for instance:  https://github.com/postmates/cernan/blob/b2ef685e5e28c83613a68a4e2f4d699376df0480/src/sink/prometheus.rs#L344) and that's defined [here](https://github.com/postmates/cernan/blob/b2ef685e5e28c83613a68a4e2f4d699376df0480/src/metric/telemetry.rs#L696), so we're bouncing out to CKMS' sum: https://github.com/postmates/quantiles/blob/1f4232b7b968d0688c1d71f5bdfcb12c7f49a69f/src/ckms.rs#L189 The nice thing here is that sum is totally additive, even during AddAssign: https://github.com/postmates/quantiles/blob/1f4232b7b968d0688c1d71f5bdfcb12c7f49a69f/src/ckms.rs#L70. 

Hmm, now sample count is https://github.com/postmates/cernan/blob/b2ef685e5e28c83613a68a4e2f4d699376df0480/src/metric/telemetry.rs#L174 which for SUMMARIZE bounces to https://github.com/postmates/quantiles/blob/1f4232b7b968d0688c1d71f5bdfcb12c7f49a69f/src/ckms.rs#L337 which is also only added to, so we're golden. 

What'll be a little weird, since we're going to make the samples happen over a time window is we'll need to keep track of previous windows' sum and count. Which, easily done, just need to watch out for that in review.

blt: @benley 671841013968cad301430d8bf257a6b6eed9c507 + fbc9baaba30b2ff6e53dc1ec3372ece1276e09fe + bba51840e32f3e96e1d379c3318712733e0fd6ee ought to be what we're looking for.

blt:
@benley 1a964729fe1f3ee4d88981f019bfa8ab4fee3cd7 pulls in the 0.6.1 release of quantiles, repairing the AddAssign issue you found.

blt:
@benley with the relaxed vector in place at 166af75bbb5d57c9bc6e89b5affa2c047e32f91c I modified your script to just send to cernan and found that larger payloads would tend to get truncated, spread across multiple packets. Something around 25 lines per payload worked fine, fitting inside whatever buffer OSX is filling. 

Confirmed your results, with regard to single-line payloads over statsd. Sometimes you get all 250, sometimes you get much less. The TCP protocols don't have that issue, so I'd say we're looking at UDP rather than something native to the prometheus sink implementation.

benley:
Cool, that confirms what I was thinking as well.

PR #310, opened 2017-09-09
Drop deleted files from the file source

### LINKED ISSUE #244: Cernan tracking deleted files
Here's an lsof output excerpt:
```
COMMAND   PID USER   FD   TYPE  DEVICE   SIZE/OFF     NODE NAME
cernan     3877  3912         root   19r      REG              202,1     337348     149000 /var/log/upstart/consul-agent.log.1 (deleted)
cernan     3877  3912         root   34r      REG              202,1        398     149073 /var/log/upstart/oxingest-common.log.1 (deleted)
cernan     3877  3912         root   35r      REG              202,1 1445977984     149017 /var/log/upstart/viper.log.1 (deleted)
cernan     3877  3912         root   41r      REG              202,1        117     149198 /var/log/upstart/cernan.log.1 (deleted)```

blt:
Prior to this commit the file source has assumed that descriptors
to deleted files can be cleared out when a new file is dropped in
the path. This is a problem for clients where log files are made
only at boot and then deleted. We now clear files in the same way
we used too but also when NotFound is returned from io operations.

This resolves #244


blt:
This is not quite right – `reset_from_md` needs to be fiddled with to manage the errors so we do read when possible – but it's close.

blt:
Oh, wait, I have an idea.

blt:
Okey dokey, pretty sure I got it. @dvdklnr confirmed that this implementation drops file descriptors pretty quickly, picks up the next file etc etc. 

Slightly lower CPU load too, with a trade-off for potentially increased read latency in systems where logs are infrequent.

tsantero:
why do we change `max_read_lines` to `max_lines_read` in the `FileServerConfig`?

blt:
@tsantero I kept the existing name in configuration but found myself getting confused by the meaning of `max_read_lines` when applied in the fileserver.

tsantero:
@blt yeah, i saw it's the same in the config (+1 for not introducing a breaking change in the config format). and yeah, `max_lines_read` feels a more natural/logical. the change _did_ however stand out, so i felt it was worth at least mentioning during review.

PR #320, opened 2017-09-22
Allow histograms to be flipped on for statsd metrics

### LINKED ISSUE #307: Support HISTOGRAM in statsd source
As discussed in #290 we want to expose using HISTOGRAM as the basis for statsd histogram / timer. The discussion in that issue didn't really come to a consensus on how we want to achieve this. Of note: 

  * individual cernans cannot dynamically choose bin totals / widths (doing so breaks aggregation in multi-cernan environments)
  * extending the protocol to add bin totals / widths is a real pain 

My inclination is to follow the lead of etsy/statsd and allow the statsd source to be configured with: 

  * a generic, one-size-fits-all configuration 
  * specific configuration decided by metric name

When a new line comes through the source, the specific configuration is used if the name hits, else we use the generic. 

@bitglue @tfheen @tsantero any thoughts / concerns?

blt:
This commit adds a 'mapping' capability for statsd source. It looks
like so:

    [sources]
      [sources.statsd.primary]
      enabled = true
      host = "localhost"
      port = 1024
      forwards = ["sinks.console", "sinks.null"]

      [sources.statsd.primary.mapping]
      [sources.statsd.primary.mapping.foo]
      mask = "foo.*"
      bounds = [0.0, 1.0, 10.0]

      [sources.statsd.primary.mapping.bar]
      mask = "bar.*"
      bounds = [0.0, 2.0, 20.0]

      [sources.statsd.primary.mapping.default]
      mask = ".*"
      bounds = [0.0, 0.1, 10.0, 100.0]

In this example three mappings are created: 'foo', 'bar' and 'default'.
The 'foo' mapping will match those metrics whose names are 'foo.*' where
the glob is greedy, 'bar' matches 'bar.*' and default catches all that
remain. In this configuration no statsd timers or histograms will not
be an AggregationMethod::Histogram. If two mappings should overlap in
their masks the one which appears first in the configuration will be
the preferred mapping.

Resolves #307

### DIFF
--- a/src/protocols/statsd.rs
+++ b/src/protocols/statsd.rs
@@ -61,56 +61,59 @@ pub fn parse_statsd(
                             return false;
                         };
                         metric = match (&src[offset..]).find('@') {
-                            Some(sample_idx) => {
-                                match &src[offset..(offset + sample_idx)] {
-                                    "g|" | "g" => {
-                                        let sample = match f64::from_str(
-                                            &src[(offset + sample_idx + 1)..],
-                                        ) {
-                                            Ok(f) => f,
-                                            Err(_) => return false,
-                                        };
-                                        metric = metric.persist(true);
-                                        metric = if signed {
-                                            metric.kind(AggregationMethod::Sum)
-                                        } else {
-                                            metric.kind(AggregationMethod::Set)
-                                        };
-                                        metric.value(val * (1.0 / sample))
-                                    }
-                                    "c|" | "c" => {
-                                        let sample = match f64::from_str(
-                                            &src[(offset + sample_idx + 1)..],
-                                        ) {
-                                            Ok(f) => f,
-                                            Err(_) => return false,
-                                        };
-                                        metric = metric
-                                            .kind(AggregationMethod::Sum)
-                                            .persist(false);
-                                        metric.value(val * (1.0 / sample))
-                                    }
-                                    "ms" | "ms|" | "h" | "h|" => {
-                                        let sample = match f64::from_str(
-                                            &src[(offset + sample_idx + 1)..],
-                                        ) {
-                                            Ok(f) => f,
-                                            Err(_) => return false,
-                                        };
-                                        metric = metric.persist(false);
-                                        metric = metric
-                                            .kind(AggregationMethod::Summarize);
-                                        for &(ref mask_re, ref bounds) in config.histogram_masks.iter() {
-                                            if mask_re.is_match(name) {
-                                                metric = metric.kind(AggregationMethod::Histogram).bounds(bounds.clone());
-                                                break;
-                                            }
-                                        }
-                                        metric.value(val * (1.0 / sample))
-                                    }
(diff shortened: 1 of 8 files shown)


blakebarnett:
doh, have to add support for this in helm template

blt:
@blakebarnett ah crap, I didn't even think to ping you about this. Apologies.

src/protocols/statsd.rs
```diff
@@ -325,11 +344,13 @@ mod tests {
...
     fn test_counter() {
         let metric = sync::Arc::new(Some(Telemetry::default()));
+        let config = sync::Arc::new(StatsdParseConfig::default());
         let mut res = Vec::new();
         assert!(parse_statsd(
             "a.b:3.1|c\na-b:4|c|@0.1\na-b:5.2|c@0.2\n",
             &mut res,
-            metric
+            metric,
+                config,
```
benley: indentation looks like it might be off here

src/source/statsd.rs
```diff
@@ -25,10 +26,35 @@ pub struct Statsd {
...
+    parse_config: sync::Arc<StatsdParseConfig>,
+}
+
+/// The mask type for metrics in `StatsdParseConfig`.
+pub type Mask = Regex;
+
+/// The bound type for metrics in `StatsdParseConfig`.
+pub type Bounds = Vec<f64>;
+
+/// Configuratoin for the statsd parser
```
benley: s/Configuratoin/Configuration/

src/source/statsd.rs
```diff
@@ -25,10 +26,35 @@ pub struct Statsd {
...
+#[derive(Debug, Clone)]
+pub struct StatsdParseConfig {
+    /// Set specific bin masks for timeseries according to their name. The name
+    /// may be a globbing match, such like 'foo.*'. In this case all metrics
```
benley: Might want to say `regex match` instead of globbing here, since it's not using typical wildcard globbing semantics (afaict)

benley:
In the example from the PR description, you have:

```
  [sources.statsd.primary.mapping.default]
  mask = "*"
  bounds = [0.0, 0.1, 10.0, 100.0]
```

Will that mask actually work, or would that need to be `.*`?

blt:
@benley good catch, that's a typo. I've corrected it. Result of typoed version:

```
➜  cernan git:(histogram_switch_for_statsd) ✗ cargo run -- --config examples/configs/dev.toml
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/cernan --config examples/configs/dev.toml`
thread 'main' panicked at 'mask is not a valid regex: Syntax("Error parsing regex near \'*\' at character offset 0: Missing expression for repetition operator.")', src/libcore/result.rs:906:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

benley:
OK cool, that means that at least I somewhat grasp what's going on in the code :-)

PR #322, opened 2017-09-26
Add HISTOGRAM support to the wavefront sink

### LINKED ISSUE #308: Support HISTOGRAM in wavefront sink
As of this writing the wavefront sink cannot properly ship histograms, in fact they are ignored. If 
 #307 is to land we're going to need to correct that. We can either:

  * ship the bins + count information or 
  * estimate quantiles from histogram, ship estimate 

I'm not so sure of the wavefront query language, if users would be able to construct a quantile estimate from shipped bins + count. Open to suggestions. 

@bitglue @pulltab

blt:
This commit adds emission of the histogram summarization to the
wavefront sink. The name of the metric is smushed together with
the bin boundary and the count / timestamp happen where you'd
expect.

The bin is separated by '_' rather than a dot to avoid confusion
with bins that have a decimal value.

Resolves #308. Related to #307.

### DIFF
--- a/src/sink/wavefront.rs
+++ b/src/sink/wavefront.rs
@@ -483,9 +483,32 @@ impl Wavefront {
     ) -> () {
         let mut tag_buf = String::with_capacity(1_024);
         match value.kind() {
-            AggregationMethod::Histogram => {
-                unimplemented!();
-            }
+            AggregationMethod::Histogram => if let Some(bins) = value.bins() {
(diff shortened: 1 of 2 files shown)


blt:
Example output: 

```foo_0 0 1506452465
foo_0.1 0 1506452465
foo_10 1 1506452465
foo_100 0 1506452465
foo_pos_inf 0 1506452465
```

PR #327, opened 2017-09-30
Resolve Prometheus backup issue

blt:
In high cardinality situations we observed that the prometheus sink
could back up into its queue. This was a result of the valve_state
being in place for a sink for which it is not appropriate. Sink
telemetry was misleading as the self-telemetry is only updated
during an insertion, which could not happen once valve was locked
shut. The valve state would never clear because prometheus does
not drop timeseries.

This sink also resolves a drop issue with the sink run loop in
some cases and removes the valve state concept from the filter as
that is only complicated and has no real value.

Cernan should now be more robust to high-cardinality timeseries
but the prometheus sink is susceptible to memory attacks if
random names are pumped through.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -54,10 +54,6 @@ pub trait Filter {
         res: &mut Vec<metric::Event>,
     ) -> Result<(), FilterError>;
 
-    /// Lookup the `Filter` valve state. See `Valve` documentation for more
-    /// information.
-    fn valve_state(&self) -> util::Valve;
-
     /// Run the Filter
     ///
     /// It is not expected that most Filters will re-implement this. If this is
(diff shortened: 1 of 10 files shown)


src/filter/mod.rs
```diff
@@ -77,23 +73,17 @@ pub trait Filter {
...
-                            },
-                            Err(fe) => {
-                                error!(
-                                    "Failed to run filter with error: {:?}",
-                                    name_in_fe(&fe)
-                                );
-                                let event = event_in_fe(fe);
-                                util::send(&mut chans, event);
-                            }
+                    match self.process(event, &mut events) {
```
pulltab: From review:  We should decrement attempts here.

src/sink/mod.rs
```diff
@@ -54,41 +57,63 @@ pub trait Sink {
         let mut attempts = 0;
         let mut recv = recv.into_iter();
         let mut last_flush_idx = 0;
+        // The run-loop of a sink is two nested loops. The outer loop pulls a
+        // value from the hopper queue. If that value is Some the inner loop
+        // tries to do something with it, only discarding it at such time as
+        // it's been delivered to the Source.
```
pulltab: Source ~> Sink

src/filter/mod.rs
blt: I was wrong when we were discussing this. Line 75 has a reset to 0 in the success branch.

PR #335, opened 2017-11-03
Remove windowing for Summarize in prometheus sink

blt:
Per @benley it doesn't make good sense for the prometheus sink to
hold everything perpetually save summarize in a retain_limit window.
The windowing is now removed and summarize will act as the other
aggregations do.

### DIFF
--- a/src/sink/prometheus.rs
+++ b/src/sink/prometheus.rs
@@ -8,9 +8,7 @@
 //!   - QUANTILES -> summary
 //!   - HISTOGRAM -> histogram
 //!
-//! All points are retained in a sliding window defined by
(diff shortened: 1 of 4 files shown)


src/sink/prometheus.rs
```diff
@@ -8,9 +8,7 @@
 //!   - QUANTILES -> summary
 //!   - HISTOGRAM -> histogram
 //!
-//! All points are retained in a sliding window defined by
-//! `retain_limit`. Points will persist indefinately unless new points come
-//! behind them to push them out of the `retain_limit` window.
+//! All points are retained indefinately in their aggregation.
```
benley: nit: s/indefinately/indefinitely/

PR #337, opened 2017-11-03
Prometheus source (WIP)

### LINKED ISSUE #336: Cernan requires a Prometheus Source
For systems that have gone all-in on Prometheus but need to bridge to other kinds of sinks cernan should be able to read from Prometheus.

benley:
Introduces a Prometheus source into cernan. 

Resolves #336


blt:
@benley as noted in the commit message for b51bf1c1c9b5e83838c5ad72e1e2b6f0947c8ca8 we have a little problem with ingesting prometheus "summary": we lose the underlying points when prometheus exports a summary and can't build a cernan Summarize aggregation. Not sure yet what to do about that – seems like a big detail – and would welcome any thoughts there.

blt:
I started knocking together a parser for prometheus text exposition format here: https://github.com/blt/prometheus_parsing_test It's about the level of difficulty I figured it'd be: goofy but not long-term work. 

@tsantero let me know if there's a strong ask for Prometheus protocol -> wavefront and I'll keep at this. Elsewise I'll probably do it in odd hours as hobby work.

blt:
Absent any work on this PR I'm going to go ahead and close it out. Would be very happy to see it picked back up, that said.

PR #338, opened 2017-11-06
Remove any unwrap / excepts in prometheus critical path

blt:
It turns out it was possible for the far-side of the prometheus
sink to hang up during a binary write and cause the sink to go
sideways. This commit removes that possibility.

We also do not need to recombine any longer as all points are
'perpetual'.

Signed-off-by: Brian L. Troutwine <brian@troutwine.us>


blt:
@lynchc relevant to your interests

Issue #34, opened 2016-07-13
Handle Signals 


blt:
It would be very useful for integration testing if SIGINT were to cause backends to flush and then result in program exit.

pulltab:
Another use case for signal handling: being able to promise at least once delivery on certain sinks.  

For example, consider a scenario where we want to ensure a sink delivers log messages at least once.  As an administrator, I can ensure tooling exists to bring down the system gracefully, however,  said tooling needs to inform Cernan to both shut down source ingestion and synchronously flush the configured sinks.  Making use of signals to accomplish this represents the path of least-astonishment.

pulltab:
First step - #333

blt:
I'd say #339 fit the bill!

PR #339, opened 2017-11-10
Introducing Graceful Shutdown

pulltab:
Sources, sinks, and filters now shutdown gracefully and in a manner which fosters at least once delivery.

In response to a SIGINT or SIGTERM, the cernan main thread now:

1) Gracefully terminates and joins all running sources.   For sources with protocols enabled with ACKs, this will ensure downstream clients are aware when their intended source is no longer available.  Graceful terminate is achieved by use of mio and its polling features to select between client traffic and so called `SYSTEM` events.  At present there is only one supported SYSTEM event (SHUTDOWN). 

2) Gracefully terminates running filters and sinks, blocking until all have terminated.  This is accomplished by injecting the new `metric::event::Shutdown` at the end of Hopper queue associated with the given consumer.  Upon receipt of a `metric::event::Shutdown`, filters and sinks flush and terminate.


pulltab:
Resolves #34

blakebarnett:
This is awesome, not proficient enough with rust to review properly but I can't wait to see it in action :)

src/util.rs
```diff
@@ -1,5 +1,7 @@
 //! Utility module, a grab-bag of functionality
 
+extern crate mio;
```
pulltab: This needs to be cleaned up.

src/bin/cernan.rs
```diff
@@ -50,6 +65,26 @@ fn populate_forwards(
...
 
+fn join_all(workers : HashMap<String, SinkWorker>){
+    for (_worker_id, worker) in workers {
+        worker.thread.join().expect("Failed to join worker");
+    }
+}
+
+fn broadcast_shutdown(workers : &HashMap<String, SinkWorker>){
+    let mut source_channels = Vec::new();
+    for (id, worker) in workers.iter() {
```
pulltab: @blt This broadcast approach thwarts at least once delivery. 

Consider a case where we have a topology where a sink is at the end of some chain of sinks/filters.  What we really want is to inject the SHUTDOWN at the top most sink/filter for each chain then let that SHUTDOWN event be forwarded along.

Do you happen to know a way to iterate over the root nodes of each chain?

blt:
Right now it’s goofy because there’s no proper graph structure but we build
up the “top” of the paths for time pulses. I imagine you could reuse that
list.

Somewhat longer term we should probably invest in a proper graph. Cycle
detection would be nice.

src/bin/cernan.rs
blt: Right now it’s goofy because there’s no proper graph structure but we build
up the “top” of the paths for time pulses. I imagine you could reuse that
list.

Somewhat longer term we should probably invest in a proper graph. Cycle
detection would be nice.

src/source/internal.rs
```diff
@@ -106,312 +108,321 @@ macro_rules! atom_non_zero_telem {
...
-                );
-                atom_non_zero_telem!(
-                    "cernan.filters.delay.log.accept",
-                    filter::delay_filter::DELAY_LOG_ACCEPT,
-                    self.tags,
-                    self.chans
-                );
-                while let Some(mut telem) = Q.pop() {
+            let mut events = mio::Events::with_capacity(1024);
+            match poll.poll(& mut events, Some(slp)) {
```
blt: `poll.poll` will block, will it not? If it does, that's a problem for the steady beat of telemetry. 

If we're not catching a shutdown here should we be polling?

blt:
Hhhmmmm so this works fine in debug mode, but as soon as you compile in release-mode sending a SIGKILL will freeze things up.

blt:
No, no, I can trigger it in debug mode too. Never mind my last.

src/sink/prometheus.rs
```diff
@@ -572,6 +572,14 @@ impl Sink for Prometheus {
         // nothing, intentionally
     }
 
+    fn shutdown(&mut self) -> () {
+        self.flush();
+        // This won't work:
+        // https://docs.rs/hyper/0.10.13/hyper/server/struct.Listening.html#method.close
```
blt: @pulltab here's the problem.

pulltab: Does the following line you introduced resolve the issue?

src/source/internal.rs
pulltab: It does block, with the same slp value. 

Same behavior as https://github.com/postmates/cernan/pull/339/files#diff-fa755baccca5ebdb28f49f30a86be0f7L112.

My expectation is that one day we will be catching system event here, hence it was plumbed in.

src/sink/prometheus.rs
blt: No. :(

src/source/graphite.rs
```diff
@@ -64,87 +68,165 @@ impl Graphite {
...
+                }
+            },
+        }
+    }
+}
+
+fn handle_tcp(
+    chans: util::Channel,
+    tags: sync::Arc<metric::TagMap>,
+    socket_map: HashMap<mio::Token, mio::net::TcpListener>,
```
blt: Oops, `HashMap` in the wild. Hoping to convert this to slab.

src/source/native.rs
```diff
@@ -70,122 +76,206 @@ impl NativeServer {
...
-                    "new peer at {:?} | local addr for peer {:?}",
-                    stream.peer_addr(),
-                    stream.local_addr()
-                );
-                let tags = tags.clone();
-                let chans = chans.clone();
-                thread::spawn(move || {
-                    handle_stream(chans, tags, stream);
+    tags: sync::Arc<metric::TagMap>,
+    listener_map: HashMap<mio::Token, mio::net::TcpListener>,
```
blt: Marker for `HashMap` -> slab conversion.

pulltab: Fixed.

src/source/graphite.rs
pulltab: Fixed.

src/sink/prometheus.rs
```diff
@@ -10,26 +10,25 @@
...
-use hyper::header::{ContentEncoding, Encoding};
-use hyper::server::{Handler, Listening, Request, Response, Server};
+use http;
 use metric;
 use metric::{AggregationMethod, TagMap};
-use protobuf::Message;
-use protobuf::repeated::RepeatedField;
-use protocols::prometheus::*;
+// use protobuf::Message;
+// use protobuf::repeated::RepeatedField;
```
blt: Marker for commented-out code.

src/sink/prometheus.rs
```diff
@@ -342,32 +339,62 @@ impl Handler for SenderHandler {
...
-                    .map(|x| x.unwrap())
-                    .collect();
-                let mut accept_proto = false;
-                for hdr in &accept {
-                    if hdr.contains("application/vnd.google.protobuf;") {
-                        accept_proto = true;
-                        break;
-                    }
-                }
+                // let default_header =
```
blt: Marker for commented-out code.

src/sink/prometheus.rs
```diff
@@ -342,32 +339,62 @@ impl Handler for SenderHandler {
...
+                    None,
+                );
+
+                match request.respond(response) {
+                    Ok(_) => {
+                        PROMETHEUS_WRITE_TEXT.fetch_add(1, Ordering::Relaxed);
+                    }
+
+                    Err(e) => {
+                        panic!(format!("Failed to send prometheus response! {:?}", e));
```
blt: We shouldn't panic here. Previously we incremented `PROMETHEUS_REPORT_ERROR` on failure.

src/sink/prometheus.rs
```diff
@@ -342,32 +339,62 @@ impl Handler for SenderHandler {
...
+                    http::StatusCode::from(200),
+                    headers.to_vec(),
+                    &buffer[..],
+                    Some(buffer.len()),
+                    None,
+                );
+
+                match request.respond(response) {
+                    Ok(_) => {
+                        PROMETHEUS_WRITE_TEXT.fetch_add(1, Ordering::Relaxed);
```
blt: Previous behaviour was to increment PROMETHEUS_WRITE_TEXT when we attempted to write, uh, text. We can probably remove both WRITE_BINARY and WRITE_TEXT telemetry now that we're all-in on text.

src/sink/prometheus.rs
```diff
@@ -377,152 +404,156 @@ impl Prometheus {
...
+    pub fn new(config: &PrometheusConfig) -> Self {
         let aggrs = PrometheusAggr::new(config.capacity_in_seconds);
         let thrd_aggrs = sync::Arc::new(sync::Mutex::new(None));
         let srv_aggrs = sync::Arc::clone(&thrd_aggrs);
-        let listener = Server::http((config.host.as_str(), config.port))
-            .unwrap()
-            .handle_threads(SenderHandler { aggr: srv_aggrs }, 1)
-            .unwrap();
 
+        let host_port =
```
blt: A tuple here should be able to coerce into the IP needed. Yeah, looks like `(&str, u16)` implements [ToSocketAddrs](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html).

src/sink/prometheus.rs
```diff
@@ -377,152 +404,156 @@ impl Prometheus {
...
-                metric.set_histogram(histogram);
-                metric_family.set_field_type(MetricType::HISTOGRAM);
-                metric_family.set_metric(RepeatedField::from_vec(vec![metric]));
-                metric_family.write_length_delimited_to_writer(res.by_ref())?
-            }
-        }
-    }
-    res.end()
-}
+// fn write_binary(aggrs: Iter) -> io::Result<()> {
```
blt: Marker for commented out code.

src/source/file/file_server.rs
```diff
@@ -1,20 +1,17 @@
+extern crate mio;
```
blt: Marker to remove redundant `extern crate mio` in favor of the one in `lib.rs`.

src/source/flush.rs
```diff
@@ -1,3 +1,5 @@
+extern crate mio;
```
blt: Marker to remove redundant extern crate mio in favor of the one in lib.rs.

src/source/mod.rs
```diff
@@ -3,6 +3,8 @@
 //! In cernan a `Source` is a place where all `metric::Event` come from, feeding
 //! down into the source's forwards for further processing. Statsd is a source
 //! that creates `Telemetry`, `FileServer` is a source that creates `LogLine`s.
+extern crate mio;
```
blt: Marker to remove redundant extern create.

src/source/internal.rs
```diff
@@ -1,3 +1,5 @@
+extern crate mio;
```
blt: Marker to remove redundant extern crate.

src/source/native.rs
```diff
@@ -1,14 +1,19 @@
+extern crate mio;
```
blt: Marker to remove redundant extern create.

src/source/statsd.rs
```diff
@@ -106,55 +107,79 @@ fn handle_udp(
...
+            Ok(_num_events) => for event in events {
+                match event.token() {
+                    constants::SYSTEM => return,
+                    token => {
+                        // Get the socket to receive from:
+                        let socket = &conns[token];
+
+                        let (len, _) = match socket.recv_from(&mut buf) {
+                            Ok(r) => r,
+                            Err(e) => panic!(format!(
```
blt: In the event that we can't read from the socket here, what ought we to do? Is this a catastrophic failure – as indicated by the panic? – or something that can be recovered from? I'm not sure of the answer.

src/thread.rs
```diff
@@ -0,0 +1,74 @@
+//! Mio enabled threading library.
+
+extern crate mio;
```
blt: Marker for redundant extern crate.

src/sink/prometheus.rs
pulltab: Fixed.

src/sink/prometheus.rs
pulltab: Fixed.

src/sink/prometheus.rs
pulltab: Converted to a warn!

src/sink/prometheus.rs
pulltab: Removed both and added PROMETHEUS_REPORT_SUCCESS to compliment PROMETHEUS_REPORT_SUCCESS.

src/sink/prometheus.rs
pulltab: Removed.

src/source/file/file_server.rs
pulltab: Replaced with a use statement.

src/source/mod.rs
pulltab: Fixed.

src/source/internal.rs
pulltab: Removed.

src/source/native.rs
pulltab: Fixed.

src/source/statsd.rs
pulltab: It's what we did before: https://github.com/postmates/cernan/pull/339/files/61641dd333ad7f34e683ea2c86e42c68adf16612#diff-5bfbf0befe1a07571883accc30ffb6e7L119

src/thread.rs
pulltab: Fixed.

gliush:
Cool!

PR #340, opened 2017-11-20
Prometheus freeze up

blt:
This commit improves the prometheus sinks' performance. Two PRs for hopper and quantiles – referenced in the toml file – improve the baseline performance of our dependencies and we further avoid copies and string mangling in this commit. Namely:

  * sanitization is deferred until poll time, reducing contention on insertion
  * HashMap is removed in favor of two Vecs, similarly to our other sinks
  * the sink trait more eagerly reads from the hopper queue 
  * prometheus sink's 'reportable' now makes references, not clones 
  * statsd error is now 0.01 rather than 0.001

Each of these is an incremental improvement. The prometheus sinks' Summaries will continue to grow and slow, as the quantiles pull-request discusses, but the effect won't be _as_ serious so quickly.

Hopper: https://github.com/postmates/hopper/pull/18
Quantiles: https://github.com/postmates/quantiles/pull/19

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ elastic_types = "0.19"
 fern = "0.4"
 flate2 = "0.2"
 glob = "0.2.11"
-hopper = "0.3.2"
+hopper = { git = "https://github.com/postmates/hopper.git", branch = "introduce_zlo" }
 hyper = "0.10"
 hyper-native-tls = "0.2"
 lazy_static = "0.2.1"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -23,7 +23,7 @@ elastic_types = "0.19"
 fern = "0.4"
 flate2 = "0.2"
 glob = "0.2.11"
-hopper = "0.3.2"
+hopper = { git = "https://github.com/postmates/hopper.git", branch = "introduce_zlo" }
```
pulltab: This needs to be changed.

Cargo.toml
```diff
@@ -32,7 +32,7 @@ log = "0.3.6"
 lua = { git = "https://github.com/blt/rust-lua53.git", branch = "master" }
 openssl-probe = "0.1"
 protobuf = "1.2"
-quantiles = { version = "0.6", features = ["serde_support"] }
+quantiles = { git = "https://github.com/postmates/quantiles.git", branch = "ckms_insertion_perf_improvement", features = ["serde_support"] }
```
pulltab: Ditto here.

PR #345, opened 2017-11-25
Update elastic dependency

blt:
This commit updates the elasticsearch dependency. There's other dependency updates as well but elastic is the main show. We did have to change the internal telemetry some because the elastic API has changed but only around client errors. 

### DIFF
--- a/src/sink/elasticsearch.rs
+++ b/src/sink/elasticsearch.rs
@@ -3,7 +3,8 @@
 use chrono::DateTime;
 use chrono::naive::NaiveDateTime;
 use chrono::offset::Utc;
-use elastic::client::responses::BulkAction;
+use elastic::client::responses::bulk;
+use elastic::error;
 use elastic::error::Result;
 use elastic::prelude::*;
 use metric::{LogLine, Telemetry};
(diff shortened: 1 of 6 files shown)


src/sink/elasticsearch.rs
```diff
@@ -287,12 +253,36 @@ impl Sink for Elasticsearch {
...
+                            _ => {
                                 ELASTIC_ERROR_API_UNKNOWN
                                     .fetch_add(1, Ordering::Relaxed);
                                 error!("Unable to write, API Error (Unknown)");
                             }
                         }
+                    },
+                    error::Error::Client(ref client_error) => {
+                        ELASTIC_ERROR_CLIENT.fetch_add(1, Ordering::Relaxed);
+                        error!("Unable to write, client error: {}", client_error.description());
```
gliush: I don't see a json error handling.
Am I right saying that now we have only 2 elastic error types: Error::Api and Error::Client?

blt: @gliush yep, that's correct. The elastic library changed how it structures errors on us.

PR #347, opened 2017-11-29
Allow wavefront sink to shed points

blt:
We know that Wavefront doesn't take historical backfill data. It
is possible for the sink to back up into its buffer when the proxy
it reports to is unavailable, leading to reports that will have no
value. This option allows Ops to configure a time window to shed
points.

### DIFF
--- a/src/sink/wavefront.rs
+++ b/src/sink/wavefront.rs
@@ -62,12 +62,10 @@ pub struct Wavefront {
     aggrs: buckets::Buckets,
     delivery_attempts: u32,
     percentiles: Vec<(String, f64)>,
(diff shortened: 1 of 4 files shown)


src/sink/wavefront.rs
```diff
@@ -100,6 +98,9 @@ pub struct WavefrontConfig {
     /// Determine if we will or will not pad an aggregation, disregard ephemeral
     /// status
     pub pad_control: PadControl,
+    /// Determine the age at which a Telemetry point will be ejected. If the
+    /// value is None no points will ever be rejected.
+    pub age_threshold: Option<u64>,
```
pulltab: Can we detail the units here?

blt: Good call.

PR #349, opened 2017-12-01
Remove ES loglines when they've signaled as OK

### LINKED ISSUE #348: ElasticSearch sink losing some log lines
We are currently seeing some logs missing and notice errors like:

```cernan-g4c4t cernan [elastic::client::requests][174][2017-11-30T18:18:45.299068595+00:00][ERROR] Elasticsearch Response: correlation_id: 'be20d167-e875-4f0d-9399-b2546ddb3e81', error: 'https://search-logs55-g4fexgk6iiyjifhfmiuykc6pxu.us-west-2.es.amazonaws.com/_bulk: failed to lookup address information: Try again'
cernan-g4c4t cernan [cernan::sink::elasticsearch][243][2017-11-30T18:18:45.299097051+00:00][ERROR] Unable to write, Reqwest Client Error: failed to lookup address information: Try again
```

@blt verified that it's possible for ES to signal that the majority of items were written but not all.

blt:
It's possible for a bulk request to roll through but then for an
individual item in the request to still fail. We now store a UUID
in the ES buffer and use that UUID -- fed through to ES -- to remove
a newly stored bulk item from our buffer.

This resolves #348

### DIFF
--- a/src/sink/elasticsearch.rs
+++ b/src/sink/elasticsearch.rs
@@ -200,15 +200,10 @@ impl Sink for Elasticsearch {
                         Ok(item) => {
                             let uuid = uuid::Uuid::parse_str(item.id())
                                 .expect("catastrophic error, TID not a UUID");
-                            match self.buffer
+                            if let Ok(idx) = self.buffer
                                 .binary_search_by(|probe| probe.uuid.cmp(&uuid))
                             {
-                                Ok(idx) => {
-                                    self.buffer.remove(idx);
-                                }
-                                Err(_) => {
-                                    unreachable!();
-                                }
+                                self.buffer.remove(idx);
                             }
                             ELASTIC_RECORDS_TOTAL_DELIVERED
                                 .fetch_add(1, Ordering::Relaxed);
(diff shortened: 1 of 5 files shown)


src/sink/elasticsearch.rs
```diff
@@ -169,24 +185,40 @@ impl Sink for Elasticsearch {
...
+                            if let Ok(idx) = self.buffer
+                                .binary_search_by(|probe| probe.uuid.cmp(&uuid))
+                            {
+                                self.buffer.remove(idx);
+                            }
                             ELASTIC_RECORDS_TOTAL_DELIVERED
                                 .fetch_add(1, Ordering::Relaxed);
                         }
                         Err(item) => {
+                            let uuid = uuid::Uuid::parse_str(item.id())
```
pulltab: What not parse this uuid just before this match block?

blt: The type doesn't work out. Before this block the `item` is a `Result<Item, ?>` and here it's the item we need.

PR #358, opened 2017-12-06
remove unnecessary block

ibotty:
This is just a small detail that I noticed.

### DIFF
--- a/src/bin/cernan.rs
+++ b/src/bin/cernan.rs
@@ -247,10 +247,10 @@ fn main() {
         let internal_config = &args.internal;
         config_topology
             .insert(cfg_conf!(internal_config), internal_config.forwards.clone());
-        if let Some(ref configs) = args.statsds {
-            for (config_path, config) in configs {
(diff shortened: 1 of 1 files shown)


src/bin/cernan.rs
```diff
@@ -243,14 +243,13 @@ fn main() {
...
+    config_topology
             .insert(cfg_conf!(internal_config), internal_config.forwards.clone());
-        if let Some(ref configs) = args.statsds {
-            for (config_path, config) in configs {
-                config_topology.insert(config_path.clone(), config.forwards.clone());
-            }
+
+    if let Some(ref configs) = args.statsds {
+        for (config_path, config) in configs {
+            config_topology.insert(config_path.clone(), config.forwards.clone());
```
ibotty: Most other sources use `config_topology.insert(cfg_conf!(config), config.forwards.clone());` does it make sense to add `cfg_conf!` to any `config_topology.insert`? Just for consistency (also at sinks config above).

ibotty: Please disregard the comment. I was not reading properly.

Issue #232, opened 2017-03-23
Log Ingestion from systemd


lynchc:
Need log ingestion from systemd. something better than my crappy journalctl piping. bdb ref'd this:

https://github.com/jmesmon/rust-systemd/blob/master/tests/journal.rs

PR #359, opened 2017-12-06
RFC: systemd-journald source

ibotty:
This is getting closer to working. There are (at least) the following tasks still to do (disregarding code review requests).

- [ ]  Documentation,
- [ ]  Matching DSL in config file (selecting which messages to ingest),
- [ ]  Upstreaming rust-systemd pull-request (jmesmon/rust-systemd#35),
- [ ]  `cfg-config` it out on platforms that don't support/run systemd,
- [ ]  properly setting the timestamp,
- [ ]  tests,
- [ ]  fixing travis-ci, so it tests (at least building) this source.

Original message below.
---
This is an untested journald source.

It depends on an (still unreviewed) pull request to the systemd crate: jmesmon/rust-systemd#35

There are a few things missing. It cannot
 * filter yet what messages to ingest (by unit, or anything else);
 * ingest historical messages.

It also does not contain any documentation. Where should I put it?

I am still not sure what configuration API is best to describe matching the log records to get (systemd-journald can AND and OR matches. See `sd_journal_add_match(3)`). Or whether that is even desirable. What is easily possible and should be done before merging is simple filtering by one field.

ref #232

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,6 +193,7 @@ dependencies = [
  "serde 1.0.23 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.23 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "systemd 0.2.0",
  "tempdir 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "url 1.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -312,6 +313,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi-build 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "cstr-argument"
+version = "0.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cfg-if 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -708,6 +718,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "libsystemd-sys"
+version = "0.1.0"
+dependencies = [
+ "libc 0.2.34 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pkg-config 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "log"
 version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


blt:
This is really tidy. I think there'll be a little re-integration kerfuffle once #339 lands but hopefully not too much. The intention is for #339 to be merged in short order. 

> It also does not contain any documentation. Where should I put it?

Right now we park all documentation in the [wiki](https://github.com/postmates/cernan/wiki). Sources / Sinks are broken down in their header page according to their stability. [Sources](https://github.com/postmates/cernan/wiki/Sources). If you don't have edit privileges for the wiki do let me know. 

Every so often I look into moving cernan's docs in-tree. By the time we hit 1.0 I'd really like to be publishing on crates.io again – limited by #360 – and have in-tree docs. 

> I am still not sure what configuration API is best to describe matching the log records to get (systemd-journald can AND and OR matches. See sd_journal_add_match(3)). Or whether that is even desirable. What is easily possible and should be done before merging is simple filtering by one field.

I'd say it's desirable. I don't have a suggestion on format as yet but I'll noodle on it. Log matching is also something I'd be willing to create a follow-up issue for, on the way to stabilizing this source.

ibotty:
Thank you for your encouragement, Brian.

Could you please have a look on how to remove some `clone`s? I think
they are a bit obsessive (not in the critical path, so not necessarily
too bad, but bad style).

I will write some documentation post-merge, so the wiki does not 
describe vaporware ;).

I will add some basic matching today or tomorrow and lightly test it.

I am totally fine waiting for #339. I will rebase on top of it when it
is ready.

ibotty:
I lightly tested it today and ran into a few problems.

 * There is a bug that messages get delivered twice. I am investigating why that happens. I do think it's either in my pull request to `jmesmon/rust-systemd` or  `systemd/systemd` itself.

 * It fails in `sink::Console`'s `format_time`, because I don't pass the timestamp right. That leads me to:
How is `LogLine::time` to be used? Does it really only count the number of seconds till epoch? Only seconds resolution? I get way better resolution from journald's records (which I will continue to save in a separate field), but that sounds limiting.

When commenting out the offending lines (around `l.time()`), I get cernan to write journald messages to stdout using the console sink!

I use the following config.

```toml
[sources]
[sources.journald.one]
journal_files = "all"
runtime_only = false
local_only = true
forwards = ["sinks.console"]

[sources.journald.one.matches]
MESSAGE="aha"

[sinks]
[sinks.console]
bin_width = 10
flush_interval = 60
```

ibotty:
I think I fixed the duplicate record problem. I'll add a task list above.

Feel free to update the task list, if something comes to mind.

blt:
> Could you please have a look on how to remove some `clone`s? I think
they are a bit obsessive (not in the critical path, so not necessarily
too bad, but bad style).

Sure thing. I'm fiddling with postmates/quantiles right now but I'll be back to have a proper read-through here for sure.

> I will write some documentation post-merge, so the wiki does not
describe vaporware ;).

I appreciate that. The wiki is kinda tough in that regard.

ibotty:
#339 just got merged. I will rebase this pull request in the next days.

ibotty:
I am a little lost here. When we want to use our own event loop, we will have to reinvent quiet a bit of code because the convenient functions that libsystemd provides can't be used. Reimplementing them is mostly boring (porting modern reasonable C to rust), but I honestly can't motivate me doing it: Ultimately it is going to be not much better! The old function uses epoll (buried in some C), as will the new implementation with mio. I don't think we gain much reimplementing it.

What do you think?

blt:
@ibotty I don't think I'm following. In general, I fully support not reimplementing stuff if we can help it / the code is better for not doing so. 

Can you walk me through what would need to be reimplemented here? After @pulltab's work we've been aiming to get everything slotted nicely into mio loops.

ibotty:
Sorry. Systemd's journal API has a set of synchronous functions that are easy to work with. When using blocking, they use epoll under the hood. The asynchronous API (where you get to poll (by any means) by yourself) is way more low level.

Quote of `sd_journal_get_fd(3)` on a system with systemd's man pages:

> A synchronous alternative for using **sd_journal_get_fd()**, **sd_journal_get_events()**, **sd_journal_get_timeout()** and **sd_journal_process()** is **sd_journal_wait()**. It will synchronously wait until the journal gets changed. [...]
> Internally this call simply combines **sd_journal_get_fd()**, *sd_journal_get_events()**, **sd_journal_get_timeout()**, **poll()** and **sd_journal_process()** into one.

pulltab:
Hi, @ibotty.

From my 10 seconds of Googling it seems that `sd_journal_wait` takes a timeout parameter.  As long as your source polls for System events (read: Shutdown) on a somewhat regular basis I don't see the harm in continued use of blocking calls.

For example, consider what was done in file_server: https://github.com/postmates/cernan/blob/master/src/source/file/file_server.rs#L174

ibotty:
This is superseded by #399.

PR #365, opened 2017-12-11
Target quantiles 0.7

blt:
Quantiles 0.7 greatly improved the performance of the CKMS structure
that underpins cernan Summarize. We expect this will resolve issues
with the Prometheus sink when window sizes are non-trivial or the
number of points flowing into cernan is non-trivial.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,12 +97,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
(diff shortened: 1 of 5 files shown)


blt:
This commit exposes an 'age_threshold' similar to the wavefront sink. I don't intend to document this yet, intending to keep it back just for private fiddling.

PR #370, opened 2017-12-15
Fix compilation warning in Prometheus sink

kichristensen:
There is no need for buffer to be mutable


blt:
Thanks @kichristensen!

Issue #372, opened 2017-12-20
Internal Source With No Forwards Fails to Shutdown Cleanly


pulltab:
Configs such as the following result in a panic when SIGINT is processed by the main cernan thread:

```data-directory = "/tmp/cernan"

flush-interval = 2

[sources]

  [sources.statsd.primary]
  enabled = true
  port = 8125
  forwards = ["sinks.console"]

[sinks]
  [sinks.console]
  bin_width = 1```

Resulting panic:

```$ RUST_BACKTRACE=1 cargo run -- -vvvv --config examples/configs/barf.toml
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/cernan -vvvv --config examples/configs/barf.toml`
[cernan][111][2017-12-20T02:28:03.503826+00:00][INFO] cernan - 0.8.6
[mio::sys::unix::kqueue][95][2017-12-20T02:28:03.504467+00:00][TRACE] registering; token=Token(18446744073709551615); interests=Readable
[mio::sys::unix::kqueue][95][2017-12-20T02:28:03.504569+00:00][TRACE] registering; token=Token(18446744073709551615); interests=Readable
[cernan][48][2017-12-20T02:28:03.504609+00:00][TRACE] Populating sender from "sources.statsd.primary" to "sinks.console"
[mio::poll][784][2017-12-20T02:28:03.504578+00:00][TRACE] registering with poller
[mio::poll][784][2017-12-20T02:28:03.504777+00:00][TRACE] registering with poller
[mio::sys::unix::kqueue][95][2017-12-20T02:28:03.504896+00:00][TRACE] registering; token=Token(18446744073709551615); interests=Readable
[mio::poll][784][2017-12-20T02:28:03.506795+00:00][TRACE] registering with poller
[mio::sys::unix::kqueue][95][2017-12-20T02:28:03.506826+00:00][TRACE] registering; token=Token(0); interests=Readable
[mio::poll][784][2017-12-20T02:28:03.506858+00:00][TRACE] registering with poller
[mio::sys::unix::kqueue][95][2017-12-20T02:28:03.506886+00:00][TRACE] registering; token=Token(1); interests=Readable
[cernan::source::statsd][185][2017-12-20T02:28:03.507034+00:00][INFO] server started on *:8125
^C[cernan][773][2017-12-20T02:28:04.347590+00:00][INFO] Signaling shutdown to "sources.internal"
[...]

PR #373, opened 2017-12-20
Alters util::send to NoOp When Chans Is Empty

pulltab:
Supports use cases where sources have no configured forwards but attempt
to emit an event (such as Shutdown).


pulltab:
Resolves #372

PR #377, opened 2017-12-22
Introduces Baseline Avro Source

pulltab:
New source accepts length prefixed (32 bits, big endian) avro payload.
Such payloads are wrapped as a new event type, metric::Event::Raw which
captures the payload as well as the encoding.

This mod also introduces a generic TCP source which can be reused by any
TCP backed source.  In addition, the Source trait has been extended to unify creation and execution of Source threads.

std::thread usage in `bin/cernan` has been replaced with `cernan:thread`.

Sinks can now consume Raw events by implementing a `deliver_raw` method.  By default, the event will simply be dropped.

### DIFF
--- a/src/bin/cernan.rs
+++ b/src/bin/cernan.rs
@@ -11,25 +11,18 @@ extern crate mio;
 extern crate log;
 extern crate openssl_probe;
 
-use cernan::constants;
 use cernan::filter::{DelayFilterConfig, Filter, FlushBoundaryFilterConfig,
                      ProgrammableFilterConfig};
 use cernan::matrix;
 use cernan::metric;
-use cernan::sink::Sink;
 use cernan::source::Source;
+use cernan::sink::Sink;
+use cernan::thread::Stoppable;
 use chrono::Utc;
 use std::collections::{HashMap, HashSet};
 use std::mem;
 use std::process;
 use std::str;
-use std::thread;
-
-#[derive(Debug)]
-struct SourceWorker {
-    thread: std::thread::JoinHandle<()>,
-    readiness: mio::SetReadiness,
-}
 
 fn populate_forwards(
     mut top_level_forwards: Option<&mut HashSet<String>>,
@@ -63,9 +56,9 @@ fn populate_forwards(
     }
 }
 
-fn join_all(workers: HashMap<String, std::thread::JoinHandle<()>>) {
+fn join_all(workers: HashMap<String, cernan::thread::ThreadHandle>) {
     for (_worker_id, worker) in workers {
-        worker.join().expect("Failed to join worker");
+        worker.join();
     }
 }
 
@@ -120,9 +113,9 @@ fn main() {
     //    after all sources have shutdown.  Shutdown events serve to bookend queued
     //    events, once read it is safe for these workers to flush any pending writes
     //    and shutdown.
-    let mut sources: HashMap<String, SourceWorker> = HashMap::new();
-    let mut sinks: HashMap<String, std::thread::JoinHandle<()>> = HashMap::new();
-    let mut filters: HashMap<String, std::thread::JoinHandle<()>> = HashMap::new();
+    let mut sources: HashMap<String, cernan::thread::ThreadHandle> = HashMap::new();
+    let mut sinks: HashMap<String, cernan::thread::ThreadHandle> = HashMap::new();
+    let mut filters: HashMap<String, cernan::thread::ThreadHandle> = HashMap::new();
     let mut senders: HashMap<String, hopper::Sender<metric::Event>> = HashMap::new();
     let mut receivers: HashMap<String, hopper::Receiver<metric::Event>> =
         HashMap::new();
@@ -363,7 +356,7 @@ fn main() {
         let sources = adjacency_matrix.pop_nodes(&config.config_path);
         sinks.insert(
             config.config_path.clone(),
-            thread::spawn(move || {
+            cernan::thread::spawn(move |_poll| {
                 cernan::sink::Null::new(&config).run(recv, sources);
             }),
         );
@@ -375,7 +368,7 @@ fn main() {
         let sources = adjacency_matrix.pop_nodes(&config.config_path.clone().unwrap());
         sinks.insert(
             config.config_path.clone().unwrap(),
-            thread::spawn(move || {
+            cernan::thread::spawn(move |_poll| {
                 cernan::sink::Console::new(&config).run(recv, sources);
             }),
         );
@@ -387,7 +380,7 @@ fn main() {
         let sources = adjacency_matrix.pop_nodes(&config.config_path.clone().unwrap());
         sinks.insert(
             config.config_path.clone().unwrap(),
-            thread::spawn(move || match cernan::sink::Wavefront::new(config) {
+            cernan::thread::spawn(move |_poll| match cernan::sink::Wavefront::new(config) {
                 Ok(mut w) => {
                     w.run(recv, sources);
                 }
@@ -405,7 +398,7 @@ fn main() {
         let sources = adjacency_matrix.pop_nodes(&config.config_path.clone().unwrap());
         sinks.insert(
             config.config_path.clone().unwrap(),
-            thread::spawn(move || {
+            cernan::thread::spawn(move |_poll| {
                 cernan::sink::Prometheus::new(&config).run(recv, sources);
             }),
         );
@@ -417,7 +410,7 @@ fn main() {
         let sources = adjacency_matrix.pop_nodes(&config.config_path.clone().unwrap());
         sinks.insert(
             config.config_path.clone().unwrap(),
-            thread::spawn(move || {
+            cernan::thread::spawn(move |_poll| {
                 cernan::sink::InfluxDB::new(&config).run(recv, sources);
             }),
         );
@@ -429,7 +422,7 @@ fn main() {
         let sources = adjacency_matrix.pop_nodes(&config.config_path.clone().unwrap());
         sinks.insert(
             config.config_path.clone().unwrap(),
-            thread::spawn(move || {
+            cernan::thread::spawn(move |_poll| {
                 cernan::sink::Native::new(config).run(recv, sources);
             }),
         );
@@ -441,7 +434,7 @@ fn main() {
         let sources = adjacency_matrix.pop_nodes(&config.config_path.clone().unwrap());
         sinks.insert(
             config.config_path.clone().unwrap(),
-            thread::spawn(move || {
+            cernan::thread::spawn(move |_poll| {
                 cernan::sink::Elasticsearch::new(config).run(recv, sources);
             }),
         );
@@ -455,7 +448,7 @@ fn main() {
                 adjacency_matrix.pop_nodes(&config.config_path.clone().unwrap());
             sinks.insert(
                 config.config_path.clone().unwrap(),
-                thread::spawn(move || {
+                cernan::thread::spawn(move |_poll| {
                     cernan::sink::Firehose::new(config).run(recv, sources);
                 }),
             );
@@ -489,7 +482,7 @@ fn main() {
             let downstream_sends = adjacency_matrix.pop_metadata(&config_path);
             filters.insert(
                 config.config_path.clone().unwrap(),
-                thread::spawn(move || {
+                cernan::thread::spawn(move |_poll| {
                     cernan::filter::ProgrammableFilter::new(c).run(
                         recv,
                         sources,
@@ -526,7 +519,7 @@ fn main() {
             let downstream_sends = adjacency_matrix.pop_metadata(&config_path);
             filters.insert(
                 config.config_path.clone().unwrap(),
-                thread::spawn(move || {
+                cernan::thread::spawn(move |_poll| {
                     cernan::filter::DelayFilter::new(&c).run(
                         recv,
                         sources,
@@ -562,7 +555,7 @@ fn main() {
             let downstream_sends = adjacency_matrix.pop_metadata(&config_path);
             filters.insert(
                 config.config_path.clone().unwrap(),
-                thread::spawn(move || {
+                cernan::thread::spawn(move |_poll| {
                     cernan::filter::FlushBoundaryFilter::new(&c).run(
                         recv,
                         sources,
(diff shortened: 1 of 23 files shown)


src/metric/event.rs
```diff
@@ -1,6 +1,13 @@
 use metric::{LogLine, Telemetry};
 use std::sync;
 
+/// Supported event encodings.
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+pub enum Encoding {
```
blt: Can we also ship with an Untyped encoding?

src/metric/event.rs
```diff
@@ -22,6 +29,8 @@ pub enum Event {
     /// more events will appear.  It is expected that after receiving this
     /// marker the given source will exit cleanly.
     Shutdown,
+    /// Raw, encoded bytes.
+    Raw(Encoding, Vec<u8>),
```
blt: I will 100% forget the ordering here. How do you feel about using named fields, like: `Raw { encoding: Encoding, bytes: Vec<u8> }`? Then we can refer to them by name, match by name.

src/source/tcp.rs
```diff
@@ -0,0 +1,218 @@
...
+/// State for a TCP backed source.
+pub struct TCP {
+    config: TCPConfig,
+    thread:  thread::ThreadHandle,
+}
+
+impl Source for TCP {
+
+    fn run(&mut self, _poll: mio::Poll) {
+        //let addrs = (host_port.to_socket_addrs();
```
blt: Intentionally commented?

src/metric/event.rs
pulltab: Fixed.

src/metric/event.rs
pulltab: Added Encoding::None.

src/source/avro.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// Pulls length prefixed (4 bytes, BE), avro encoded
+/// binaries off the wire and populates them in the configured
+/// channel.
+fn handle_avro_payload(
+    mut chans: util::Channel,
+    _tags: &sync::Arc<metric::TagMap>,
+    reader: &mut io::BufReader<mio::net::TcpStream>,
+) {
+    let mut buf = Vec::with_capacity(4000);
```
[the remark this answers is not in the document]
pulltab: This was a copy and paste from other sources. 

As the buffer is resized later, I didn't think it really mattered what value was used here.

src/bin/cernan.rs
```diff
@@ -716,47 +654,35 @@ fn main() {
...
-                        "Unable to fulfill configured top-level flush to {}",
-                        destination
-                    );
-                    process::exit(0);
-                }
+    let mut flush_channels = Vec::new();
+    for destination in &flush_sends {
+        match senders.get(destination) {
+            Some(snd) => {
+                flush_channels.push(snd.clone());
```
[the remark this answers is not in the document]
pulltab: I don't know the answer to this either.  @blt can you shed some light as to why this block was performed async before?

src/metric/event.rs
pulltab: I decided to go with Encoding::Raw.

src/bin/cernan.rs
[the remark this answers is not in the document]
blt: Hmm, well, the important thing used to be https://github.com/postmates/cernan/blob/0.8.6/src/bin/cernan.rs#L530 was in a background thread so `FlushTimer` would be free to pulse down its channels. Unless I'm much mistaken this still needs to be the case, as I note down below.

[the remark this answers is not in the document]
blt: I was mistaken! The FlushTimer now creates its own thread and returns the ThreadHandle back out to us so FlushTimer makes its own self async.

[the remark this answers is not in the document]
pulltab: @blt Yes, the FlushTimer now spawns its own thread.

I believe @gliush is referring to this line which previously was executed by the FlushTimer thread but is now being executed by the main thread:

```
    let mut flush_channels = Vec::new();
    for destination in &flush_sends {
        match senders.get(destination) {
            Some(snd) => {
                flush_channels.push(snd.clone());
            }
            None => {
                error!(
                    "Unable to fulfill configured top-level flush to {}",
                    destination
                );
                process::exit(0);
            }
        }
    }
```

As this check is fatal and we cannot do much without a FlushTimer thread, I believe it is appropriate to validate that we have all the senders we need in the main thread.

src/source/avro.rs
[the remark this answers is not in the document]
dparton: It doesn't look like you use the `buf` until we know the size it needs to be. Perhaps you can just read the payload size off, and then create the buf with the required capacity, then resize it and read into it, to prevent a potential reallocation.

src/source/avro.rs
```diff
@@ -0,0 +1,85 @@
...
+            trace!("Failed to deserialize container: {:?}", e.description());
+            return;
+        }
+    }
+
+    util::send(&mut chans, metric::Event::Raw{encoding:metric::Encoding::Avro, bytes: buf});
+}
+
+impl source::Source<source::TCPConfig> for Avro {
+    /// Creates and starts an Avro source witht the given config.
```
dparton: Minor comment typo here `witht`

src/source/file/file_server.rs
```diff
@@ -56,15 +55,106 @@ impl Default for FileServerConfig {
...
+            let lim = backoff_cap.saturating_mul(2);
+            if lim > 2_048 {
+                backoff_cap = 2_048;
+            } else {
+                backoff_cap = lim;
+            }
+        } else {
+            backoff_cap = 1;
+        }
+        let backoff = backoff_cap.saturating_sub(global_lines_read);
```
dparton: What's the idea here? Seems to me this value will always either be `backoff_cap` or `0`:
* if global_lines_read > 0, then backoff_cap is 1. saturating_sub will produce 0.
* if global_lines_read = 0, then the subtraction does nothing.

pulltab: This is a block of code which was moved. 

@blt Can you please provide feedback to the original intent of this line?

src/source/avro.rs
pulltab: Fixed and updated the comment to be accurate.

src/source/avro.rs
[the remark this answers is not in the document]
pulltab: Solid point.

I moved the buffer creation to after reading the length off the wire.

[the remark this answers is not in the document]
blt: For this to avoid potential re-allocations the buf will have to live outside of `handle_avro_payload`. Right now we're allocating a buf of 4000 each time we enter the function, then resizing it.

src/source/avro.rs
blt: I still see `witht`?

src/source/avro.rs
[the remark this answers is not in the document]
blt: Ah, but you can't do that anyhow since that buffer is going to be owned by the event. Never mind.

src/source/avro.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    let mut buf = Vec::with_capacity(payload_size_in_bytes);
+    if reader.read_exact(&mut buf).is_err() {
+        return;
+    }
+
+    match serde_avro::de::Deserializer::from_container(&buf[..]) {
+        Ok(_avro_de) => {
+            trace!("Successfully deserialized container.");
+            // TODO - Enforce configurable type naming requirements.
```
blt: Outstanding todo here yet. Okay?

src/source/file/file_server.rs
```diff
@@ -56,15 +55,106 @@ impl Default for FileServerConfig {
...
-impl FileServer {
-    /// Make a FileServer
-    ///
-    pub fn new(chans: util::Channel, config: FileServerConfig) -> FileServer {
-        FileServer {
-            chans: chans,
-            pattern: config.path.expect("must specify a 'path' for FileServer"),
-            tags: config.tags,
-            max_lines_read: config.max_lines_read,
+fn file_server(mut chans: util::Channel, config: FileServerConfig, poller: mio::Poll) {
```
blt: This... makes me nervous. Why are we getting rid of trait implementations here?

src/source/avro.rs
pulltab: Leaving a note for a future feature add.

serde_avro doesn't support getting at the schema name just yet. As such, this feature can't be added now.

src/source/file/file_server.rs
pulltab: New have been moved into the Source trait.

src/source/flush.rs
```diff
@@ -1,43 +1,49 @@
...
+/// The source of all flush pulses.
 pub struct FlushTimer {
     chans: util::Channel,
 }
 
-impl FlushTimer {
+/// Nil config for FlushTimer.
+pub struct FlushTimerConfig;
+
+fn tick_tock(mut chans: util::Channel, _poll: mio::Poll) {
```
blt: Okay, I guess I see what's going on. Rather than have a single function that we pass into a thread – jettisoning the impl in the process – I'd rather see a function on the impl that gets passed into a thread.

pulltab: Which impl?  Source?

src/source/internal.rs
```diff
@@ -99,335 +88,350 @@ macro_rules! atom_non_zero_telem {
     }
 }
 
+fn handle_internal_metrics(
```
blt: This for sure follows the pattern of the original graphite source but that's my fault: a bare function hanging out in the private space of a module is untidy. Each one of these `handle...` functions look to have similar – or the same? – arguments. Can we make `handle` part of the `Source` trait?

src/source/flush.rs
blt: Yeah, probably Source. The appeal, to me, of encoding all this into traits – rather than mostly similar functions – is that it should make cranking out a new source / sink / filter straightforward. Plus, the compiler checks up on you.

src/source/file/file_server.rs
blt: Okay. I guess what I'm getting at is I'd like to see `file_server` as part of some standard interface. Maybe a `handle` on Source?

src/source/internal.rs
pulltab: Untidy is a subjective notion.

We may be able to force `handle` to be in the `Source` trait, however it is going to carry with it the "baggage" of `Send + Sync + Clone` on the state struct for the given source.  For sources this may be ok, as they don't tend to us a lot of external packages to ingest data.

src/source/file/file_server.rs
blt: Huh, yeah, seems like some dodgy math. Good catch @dp.

src/source/flush.rs
pulltab: Lets consolidate this discussion in the thread in internal.rs.

src/source/file/file_server.rs
pulltab: Ja, let's continue to discuss `handle` in internal.rs's thread.

src/source/internal.rs
blt: It is a subjective notion. I would rather see Source follow the lead of Sink / Filter and be defined in terms of standard interfaces. 

To be clear, the `Send + Sync` constraint exist because we've made the Source responsible for running its own self in a thread where it does not exist for Sink / Filter because bin/cernan.rs runs them in a thread?

pulltab: Right.

I hit this case in https://github.com/postmates/cernan/pull/378 when moving the threading model into the Sink interface.

I will attempt to introduce a `handle` callback to `Source` now and see how much code catches fire.

blt: I'll take a look at #378.

pulltab: Most of the action is here - https://github.com/postmates/cernan/pull/378/files#diff-a9892b62ac2d6b57acde6b4ee3d0fb53R46

pulltab: See https://github.com/postmates/cernan/pull/377/commits/344c50ec3cd380dddd60035c6889a063597a8070

src/source/file/file_server.rs
```diff
@@ -21,22 +22,21 @@ use util::send;
...
 pub struct FileServer {
-    chans: util::Channel,
     pattern: PathBuf,
     max_lines_read: usize,
-    tags: metric::TagMap,
 }
 
 /// The configuration struct for `FileServer`.
-#[derive(Debug, Deserialize)]
+#[derive(Clone, Debug, Deserialize)]
```
blt: Now that we're not storing the config in RunnableSource do we still need these clones? Not a problem if we do, more just curious.

pulltab: We don't technically use this constraint, AFAIK.

Reasons why I left them in:

1) Like Debug, Clone is real handy to have when you need it.

2) This constraint reinforces my notion that configs should contain simple types, leaving the transformation from config to more complex types up to the sources that consume them.

src/source/native.rs
```diff
@@ -57,221 +47,130 @@ impl Default for NativeServerConfig {
...
-    ) -> NativeServer {
-        NativeServer {
-            chans: chans,
-            ip: config.ip,
-            port: config.port,
-            tags: config.tags,
+impl From<NativeServerConfig> for TCPConfig {
+    fn from(item: NativeServerConfig) -> Self {
+        TCPConfig {
+            host: item.ip,
```
pulltab: @blt Technically speaking there is no difference (short of some field names) between NativeServerConfig/GraphiteConfig and TCPConfig.

Should we go through the exercise of porting these types now?

blt: Hmm, you know, that's a good question. I don't have an answer but I bet we'll know with hindsight. Maybe let's make an issue for 1.0 and see how we feel then?

src/source/file/file_server.rs
blt: Okay, makes sense. RE 1, agreed. RE 2, this one makes me kind of nervous since that understanding of Clone isn't enforceable by the compiler and handshake agreements in software make me nervous. But! These'll be cloned a handful of times so I'm cool with it. 

FWIW my original notion for not having these be cloneable was to statically enforce the notion that there should only be one of these in memory. Early cernan had a bug where we were cloning one of these config things each time a statsd packet rolled through. Easy to catch when you're looking for it but annoying.

blt:
@pulltab I say if this has run successfully in LLRV for a while I'm good to +1 this.

pulltab:
@blt LLRV is running along just fine on this branch.

PR #378, opened 2017-12-27
Kinesis Sink + Reworked Sink Interface

pulltab:
Bringing Sink semantics more "in line" with Source semantics:

* sink::Sink now implements a generic `new` function which generates a
  so-called RunnableSink.  Runnable exposes an API which allows
  users to consume initial state and execute the corresponding Sink as a
  thread.

* Implementing sinks are now required to implement an `init` function
  rather than a `new`.  This enables `new` to have a generic, default
  implementation which returns StatefulSink constructs.  `init` serves
  the same function as the previous `new`, namely initializing local
  state for the given sink.

* Native sink no longer implements its own run function.

A baseline Kinesis sink has been introduced which, currently, accepts only Raw events.  This new sink publishes records in batch to Kinesis, bounded by time and space.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -736,15 +736,15 @@ pub fn parse_config_file(buffer: &str, verbosity: u64) -> Args {
                     res.config_path = Some(format!("sinks.kinesis.{}", name));
 
                     let stream_name = tbl.get("stream_name").map(|x| {
-                        x.as_str().expect("stream_name must be a string")
+                        x.as_str().expect("stream_name must be a string").to_string()
                     });
 
                     if stream_name.is_none() {
+                        warn!("kinesis sink {:?} skipped as it does not provide a stream_name!", res.config_path);
                         continue;
                     };
 
-                    res.stream_name = stream_name.map(|s| s.to_string());
-
+                    res.stream_name = stream_name;
                     res.flush_interval = tbl.get("flush_interval")
                         .map(|fi| {
                             fi.as_integer().expect(
@@ -1199,6 +1199,40 @@ scripts-directory = "/foo/bar"
         assert_eq!(es.secure, true);
         assert_eq!(es.flush_interval, 2020);
         assert_eq!(es.delivery_attempt_limit, 33);
+    }
+
+    #[test]
+    fn config_kinesis_sink() {
+        let config = r#"
+[sinks]
+  [sinks.kinesis]
+  stream_name = "foobar"
+  region = "us-west-2"
+  flush_interval = 100
+"#;
+
+        let args = parse_config_file(config, 4);
+
+        assert!(args.kinesises.is_some());
+        for ks in args.kinesises.unwrap() {
+            assert_eq!(ks.stream_name, Some(String::from("foobar")));
+            assert_eq!(ks.region, Region::UsWest2);
+            assert_eq!(ks.flush_interval, 100);
+        }
+    }
+
+    #[test]
+    #[should_panic]
+    fn config_kinesis_skip_when_stream_name_is_absent() {
+        let config = r#"
+[sinks]
+  [sinks.kinesis]
+  region = "us-west-2"
+  flush_interval = 100
+"#;
+
+        let args = parse_config_file(config, 4);
+        assert!(args.kinesises.is_none());
     }
 
     #[test]
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,6 +8,7 @@ use metric::TagMap;
 use rusoto_core::Region;
 use std::collections::HashMap;
 use std::env;
+use std::str::FromStr;
 use std::fs::File;
 use std::io::Read;
 use std::path::{Path, PathBuf};
@@ -22,6 +23,7 @@ use sink::ConsoleConfig;
 use sink::ElasticsearchConfig;
 use sink::FirehoseConfig;
 use sink::InfluxDBConfig;
+use sink::KinesisConfig;
 use sink::NativeConfig;
 use sink::NullConfig;
 use sink::PrometheusConfig;
@@ -99,6 +101,8 @@ pub struct Args {
     pub elasticsearch: Option<ElasticsearchConfig>,
     /// See `sinks::Firehose` for more.
     pub firehosen: Option<Vec<FirehoseConfig>>,
+    /// See `sinks::Kinesis` for more.
+    pub kinesises: Option<Vec<KinesisConfig>>,
     /// See `sources::FileServer` for more.
     pub files: Option<Vec<FileServerConfig>>,
     /// See `sources::Internal` for more.
@@ -135,6 +139,7 @@ impl Default for Args {
             native_sink_config: None,
             elasticsearch: None,
             firehosen: None,
+            kinesises: None,
             // sources
             statsds: None,
             graphites: None,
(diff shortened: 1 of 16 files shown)


src/sink/mod.rs
```diff
@@ -28,42 +29,48 @@ pub use self::null::{Null, NullConfig};
...
+impl <S, SConfig> StatefulSink <S, SConfig> where S: 'static + Send + Sync + Sink<SConfig>, SConfig: 'static + Clone + Send + Sync
+{
+
+    /// Generic constructor for StatefulSink - execution wrapper around objects implementing Sink.
+    pub fn new (recv: hopper::Receiver<Event>, sources: Vec<String>, config: SConfig) -> StatefulSink<S, SConfig>
+    {
+        StatefulSink {
+            recv: recv,
+            sources: sources,
+            config: config.clone(),
```
blt: Why does the config need to be cloned here?

src/sink/mod.rs
```diff
@@ -28,42 +29,48 @@ pub use self::null::{Null, NullConfig};
...
+/// Wrapper around a given sink implementation and its config.
+pub struct StatefulSink<S, SConfig>
+    where
+        S: Send + Sync + Sink<SConfig>,
+        SConfig: 'static + Send + Sync + Clone,
+{
+    recv: hopper::Receiver<Event>,
+    sources: Vec<String>,
+    #[allow(dead_code)] // Only here to avoid using PhantomData.
+    config: SConfig,
```
blt: Maybe I'm missing it, but I can't find a place where config is being used by StatefulSink?

If there's only the one use site you might be able to remove the clone, since only Sink would own its config.

src/sink/mod.rs
```diff
@@ -28,42 +29,48 @@ pub use self::null::{Null, NullConfig};
...
-    /// Deliver a `LogLine` to the `Sink`. Exact behaviour varies by
-    /// implementation.
-    fn deliver_line(&mut self, line: sync::Arc<Option<LogLine>>) -> ();
-    /// Deliver a 'Raw' series of encoded bytes to the sink.
-    fn deliver_raw(&mut self, _encoding: Encoding, _bytes: Vec<u8>) -> () {
-        // Not all sinks accept raw events.  By default, we do nothing.
-        return;
+
+/// Wrapper around a given sink implementation and its config.
+pub struct StatefulSink<S, SConfig>
```
blt: `StatefulSink` is public. If someone wants to make a sink should they create an instance of `Sink` or `StatefulSink`? Although I guess `StatefulSink` is public because that's the return type of `Sink::new`? 

I guess, fundamentally, I don't understand how Sink and StatefulSink are different, why the consume_loop of StatefulSink couldn't be placed inside Sink?

src/sink/mod.rs
```diff
@@ -6,64 +6,75 @@
...
+use std::marker::PhantomData;
 use std::sync;
+use thread;
 use time;
 use util::Valve;
 
-mod console;
-mod firehose;
-mod null;
+pub mod console;
```
blt: Why mark these as pub when we're re-exporting pieces of them below?

src/sink/mod.rs
```diff
@@ -6,64 +6,75 @@
...
-    fn deliver(&mut self, point: sync::Arc<Option<Telemetry>>) -> ();
-    /// Deliver a `LogLine` to the `Sink`. Exact behaviour varies by
-    /// implementation.
-    fn deliver_line(&mut self, line: sync::Arc<Option<LogLine>>) -> ();
-    /// Deliver a 'Raw' series of encoded bytes to the sink.
-    fn deliver_raw(&mut self, _encoding: Encoding, _bytes: Vec<u8>) -> () {
-        // Not all sinks accept raw events.  By default, we do nothing.
-        return;
+
+/// Wrapper around a given sink implementation and its config.
```
blt: Can you expand on what this wrapper does in the documentation? It made good sense when we talked it through but it's worth capturing that in source.

src/sink/mod.rs
pulltab: Uniformity.  The other modes are re-exported below and pub as well.

blt: I guess my question is about those, too. Originally I didn't expose these modules so we could be free to fiddle with them, choosing to export only the key types starting on line 22 / 25. 

We should choose one or the other. Either the inner modules are public or the key things are exported.

pulltab: Let's just leave them private then.  `src/source/mod.rs` leaves them private.

src/config.rs
```diff
@@ -99,6 +101,8 @@ pub struct Args {
     pub elasticsearch: Option<ElasticsearchConfig>,
     /// See `sinks::Firehose` for more.
     pub firehosen: Option<Vec<FirehoseConfig>>,
+    /// See `sinks::Kinesis` for more.
+    pub kinesises: Option<Vec<KinesisConfig>>,
```
blt: Heh, 'firehosen' and 'kinesises'.

blt: We're very good at plurals, @pulltab.

src/config.rs
```diff
@@ -718,6 +723,47 @@ pub fn parse_config_file(buffer: &str, verbosity: u64) -> Args {
             }
             firehosen
         });
+
+    args.kinesises = sinks.get("kinesis").map(|snk| {
```
blt: Could you also make a test for this in the basement of this module?

src/config.rs
pulltab: That changeset made me feel like Smeagol.

src/sink/mod.rs
pulltab: So...we can't make all of these private.

The internal source currently reaches into these various modules to do its job. :(.

src/config.rs
pulltab: Done.

src/sink/mod.rs
pulltab: Done.

src/sink/mod.rs
pulltab: Should we save making these private for a separate PR?

blt: Oh, I see. Hmm, maybe let's just put it back like it was and drop a ticket into the issue tracker for resolution, marked Help Wanted and Easy?

src/config.rs
blt: :)

blt:
@pulltab if this passes an llrv run I'm ready to +1 it.

src/sink/mod.rs
pulltab: Reverted.

src/sink/kinesis.rs
```diff
@@ -14,9 +14,19 @@ use rusoto_kinesis::{KinesisClient, PutRecordsError, PutRecordsInput,
...
-use std::sync;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use util::Valve;
 
+lazy_static! {
+    /// Total records emitted.
+    pub static ref KINESIS_PUBLISH_SUCCESS_SUM: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
+    /// Total retryable publication errors.
+    pub static ref KINESIS_PUBLISH_FAILURE_SUM: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
```
blt: How are failure and fatal distinct?

blt:
screen shot 2017-12-31 at 21 38 20

@pulltab I've just noticed that this PR is not into master but into avro-source.

pulltab:
@blt Thanks for the reminder.

I was working on this branch concurrently with the avro source.  I will likely reopen a new, cherry picked PR and link to this PR for historical purposes.

src/sink/kinesis.rs
pulltab: The comments should capture it.

Failure is retryable, fatal is not.    Fatal signals that Kinesis publication has stopped.

pulltab:
Closing this PR.  See - #379 for the rebuilt version.

PR #379, opened 2018-01-01
Introduces Kinesis Sink + Improves Sink Interface.

pulltab:
Bringing Sink semantics more "in line" with Source semantics:

* sink::Sink now implements a generic new function which generates a
so-called `RunnableSink`. `RunnableSink` exposes an API which allows
users to consume initial state and execute the corresponding Sink as a
thread.

* Implementing sinks are now required to implement an `init` function
rather than a new. This enables new to have a generic, default
implementation which returns `RunnableSink` constructs. `init` serves
the same function as the previous `new`, namely initializing local
state for the given sink.

* Native sink no longer implements its own run function.

A baseline Kinesis sink has been introduced which, currently, accepts only Raw events. This new sink publishes records in batch to Kinesis, bounded by time and space.

### DIFF
--- a/src/sink/kinesis.rs
+++ b/src/sink/kinesis.rs
@@ -94,6 +94,11 @@ impl Sink<KinesisConfig> for Kinesis {
     /// If the given record would put the buffer at capacity, then the contents
     /// are first flushed before the given record is added.
     fn deliver_raw(&mut self, _encoding: metric::Encoding, bytes: Vec<u8>) {
+        let record_too_big = bytes.len() > self.max_bytes_per_batch;
+        if record_too_big {
+            warn!("Discarding record with size {:?} as it is too large to publish!", bytes.len());
+        }
+
         let buffer_too_big = self.buffer_size + bytes.len() > self.max_bytes_per_batch;
         let buffer_too_long = self.buffer.len() > self.max_records_per_batch;;
         if buffer_too_big || buffer_too_long {
--- a/src/sink/kinesis.rs
+++ b/src/sink/kinesis.rs
@@ -1,6 +1,7 @@
 //! Kinesis sink for Raw events.
 
 extern crate base64;
+extern crate rand;
 extern crate rusoto_core;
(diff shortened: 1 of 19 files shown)


pulltab:
This PR is a rebuilt version of #378 from master instead of a feature branch.  #378 will be closed.

@blt This branch runs successfully in llrv.

pulltab:
@blt - fwiw I rebased to the latest master (with the graphite fix) and llrv is still running well.

src/sink/kinesis.rs
```diff
@@ -0,0 +1,215 @@
...
+        let encoded_bytes_len = encoded_bytes.len();
+        let record_too_big = encoded_bytes_len > self.max_bytes_per_batch;
+        if record_too_big {
+            warn!("Discarding encoded record with size {:?} as it is too large to publish!", encoded_bytes_len);
+            return;
+        }
+
+        let buffer_too_big =
+            self.buffer_size + encoded_bytes_len > self.max_bytes_per_batch;
+        let buffer_too_long = self.buffer.len() > self.max_records_per_batch;
```
pulltab: This should be >=.

src/sink/kinesis.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Encodes and records the given event into the internal buffer.
+    ///
+    /// If the given record would put the buffer at capacity, then the contents
+    /// are first flushed before the given record is added.
+    fn deliver_raw(&mut self, _encoding: metric::Encoding, bytes: Vec<u8>) {
+        let encoded_bytes = base64::encode(&bytes).into_bytes();
+        let encoded_bytes_len = encoded_bytes.len();
+        let record_too_big = encoded_bytes_len > self.max_bytes_per_batch;
+        if record_too_big {
+            warn!("Discarding encoded record with size {:?} as it is too large to publish!", encoded_bytes_len);
```
pulltab: We should add some internal telemetry here so that we can catch this case.

Issue #380, opened 2018-01-01
Graphite Source Emitting Points Very Slowly


pulltab:
After change the graphite source to a mio event loop, the stream handling logic no longer properly processes newline delimited metric blobs from upstream clients such as collectd.

Instead the graphite source only parses one metric per mio event, which translates to one metric every time a batch of events is published.

After experimentation is appears that the following holds true:

* EOF delimites batches of newline delimited metrics.

* We receive one mio readiness event per batch.

Given the above observations, it should be safe reintroduce reading metric lines in a loop until EOF is reached.

PR #381, opened 2018-01-01
Fixes Graphite Consumption Regression.

pulltab:
The mio event loops introduced since 0.8.6 brought with them a regression in
the Graphite source in which one metric is emitted per mio
readiness interval.

This change brings us back to the previous behavior wherein each EOF
delimited payload is parsed in its entirety before the Graphite source
returns to polling for System events.


pulltab:
Resolves #380

PR #382, opened 2018-01-03
Introduces OrderBy to Event::Raw

pulltab:
In some use cases, against some sinks, producers can opt to order events
by a given value which governs which partition / shard the record is
serialized within.

The wire protocol for the Avro source has been extended accordingly with
the new field being required for all payloads.  Making the field
required prompts producers (clients) to decide when they wish their
events to be serialized.


pulltab:
@blt The nightly build is failing on -

```error[E0583]: file not found for module `montgomery`
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.12.1/src/arithmetic/arithmetic.rs:15:9
   |
15 | pub mod montgomery;
   |         ^^^^^^^^^^
   |
   = help: name the file either arithmetic/montgomery.rs or arithmetic/montgomery/mod.rs inside the directory "/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.12.1/src/arithmetic"
error: aborting due to previous error
error: Could not compile `ring`.```

Have you seen this before?  It isn't happening locally.

File: src/source/avro.rs [hunk 1]
```diff
@@ -55,6 +88,19 @@ impl AvroStreamHandler {
...
+        }
+
+        let mut header_raw = Vec::with_capacity(header_size_in_bytes);
+        if reader.read_exact(&mut header_raw).is_err() {
+            warn!("Failed to read header from stream!");
+            return;
+        }
+        let header : Header = header_raw.into();
+
         let mut buf = Vec::with_capacity(payload_size_in_bytes);
```
dparton: Since the intention of the length is to have it include the header bits, and we have already read that, we should be subtracting the header size in bytes from the payload size in bytes to arrive at the remaining size to read

pulltab: Fixed.

src/filter/delay_filter.rs
```diff
@@ -80,11 +80,8 @@ impl filter::Filter for DelayFilter {
             metric::Event::TimerFlush(f) => {
                 res.push(metric::Event::TimerFlush(f));
             }
-            metric::Event::Raw { encoding, bytes } => {
-                res.push(metric::Event::Raw {
-                    encoding: encoding,
-                    bytes: bytes,
-                });
+            raw @ metric::Event::Raw {..} => {
```
blt: Oh hey, this is new syntax to me. What's this?

src/sink/kinesis.rs
```diff
@@ -111,7 +111,7 @@ impl Sink<KinesisConfig> for Kinesis {
     ///
     /// If the given record would put the buffer at capacity, then the contents
     /// are first flushed before the given record is added.
-    fn deliver_raw(&mut self, _encoding: metric::Encoding, bytes: Vec<u8>) {
+    fn deliver_raw(&mut self, order_by: u64, _encoding: metric::Encoding, bytes: Vec<u8>) {
```
blt: Hmm, at this point, why not have deliver_raw take an actual Raw, pull out what's needed in the function body rather than pass arguments?

src/source/avro.rs
```diff
@@ -55,7 +88,23 @@ impl AvroStreamHandler {
             Err(_) => return,
         };
 
-        let mut buf = Vec::with_capacity(payload_size_in_bytes);
+        let header_size_in_bytes = mem::size_of::<Header>();
+        if payload_size_in_bytes <= header_size_in_bytes {
+            warn!("Invalid payload detected!  Discarding!");
```
blt: Worth adding self-telemetry for this case. People – maybe us – will ask for it.

src/source/avro.rs
```diff
@@ -4,13 +4,36 @@ use metric;
...
-use std::{io, sync};
-use std::io::Read;
+use std::{io, mem, sync};
+use std::io::{Cursor, Read};
 use util;
 
 #[derive(Default, Debug, Clone, Deserialize)]
 pub struct AvroStreamHandler;
 
+#[repr(C, packed)]
```
blt: Uh, hmm, packed is going to be pretty crummy on chips that demand aligned memory. I also don't see where repr C is needed? 

Seems like – and maybe I'm missing something here – we don't need either of these reprs.

blt:
@pulltab seems to be a new, fun feature of nightly. There are patches landed in ring to address. I switched off to stable for now. I figure it'll get corrected for us in a day or two.

src/filter/delay_filter.rs
pulltab: Binds `raw` as an identifier and doesn't care about the contents of the enum struct.

src/source/avro.rs
pulltab: It is needed as we use `size_of` to know the size of the expected header off the wire.

src/filter/delay_filter.rs
blt: Well hell, that'll clean up other bits of cernan.

src/source/avro.rs
pulltab: Alternatively, I can just ensure that `Header` stays aligned to u64, which would require me to either add an `unused` entry or grow `version` to a u64.

blt: Oh! So Avro headers are defined as being the size of whatever C's packed representation is? Got it.

src/sink/kinesis.rs
pulltab: I had thought about it.

I was just really looking forward to a straight forward PR that didn't require me to make API changes.  :D

blt: 100% fair.

src/source/avro.rs
pulltab: Not Avro headers, but the headers we are wrapping around it, yeah.

blt: Got it.

src/source/avro.rs
dparton: I agree 💯 on having telemetry here. I'm torn on what else we should do in this case. If this happens, the wire is probably fried. Currently this just reads the length N, but doesn't then toss out/seek past the next N bytes as invalid. In the next pass, we're going to read part or all of those N bytes as the next length, and potentially have an offset issue.

What's the right thing to do then? Toss out the N bytes as encoded in the length? Disconnect the socket? Do what we do now and just keep reading from the same position?

dparton: Also: some sort of maximum payload size would be good to enforce. It's currently trivial to blow up the memory of the cernan process by sending a large length here.

blt: That'd be a good auditing project. To date we've assumed that cernan runs in a more or less trusted environment which, uh, is not even true for the environments it runs in today. One group goofing will cause a denial of service by exploding memory etc etc.

blt:
This is shaping up real nice. I say we get some self-telemetry in where the warning logs are presently and I'll be ready to approve this.

dparton:
LGTM!

PR #383, opened 2018-01-04
Updates Avro Source's Default Port to 2002

pulltab:
The new value replaces the previous 8080 and commemorates the year Buzz
Aldrin defended his honor.


blt:
Yeah, this'll all land in the wiki when it's available for public use, probably marked as experimental. 

@pulltab @gliush's second question made me think of something. config.rs isn't the appropriate place for `2002`: it should be the default of the avro config struct. That's why it was originally `unwrap_or(res.port)`.

PR #386, opened 2018-01-05
File truncation support

### LINKED ISSUE #376: Current file_watcher behaviour is confusing
The current behaviour of file_watcher is confusing. Right now we guarantee that _if_ a log file has been opened then we will read all the lines out of that file _unless_ there is truncation. But! We will only open a file by explicit path. This leads to surprising behaviour in the event of fast rotations when a log rotator is active. 

Consider this sequence of filesystem actions:

```
CREATE foo.log
(0) WRITE foo.log
(1) WRITE foo.log
(0) ROTATE foo.log => foo.log.1
(2) WRITE foo.log
(1) ROTATE foo.log => foo.log.1, foo.log.2 
```

When file_watcher reads out lines, what writes should it see? The answer now depends on when file_watcher opened which `foo.log`. If file_watcher opened `foo.log` but never read it until it became `foo.log.2` then all the writes in `foo.log.1` will be skipped. 

@pulltab has suggested that we can reduce the surprise by keeping a record of _all_ files that match the correct path and work them in order of discovery. If a file is deleted before it can be worked, we are then able to report this fact. Sounds like a plan to me.

blt:
Previously cernan worked on the assumption that log files would not be truncated and that, if they were, you were on your own. Well, it turns out that K8s + Docker regularly truncates log files and we need to support that behaviour. In this PR we do.

The main bulk of the work has been done in validation. In particular, we have two new quickcheck tests that validate the behavious of logging absent truncation -- where we can accurately predict how many writes will be read -- and in the presense of trunction -- where we cannot accurately predict but can set bounds on the number of writes that will be read.

We also emit a bit of telemetry whenever cernan detects an on-disk truncation. This will allow operators more insight into file source's view of the system proper.

Resolves #376

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ src/metrics/graphite.rs
 *.bk
 /tags
 /.vscode
+tmp/
--- a/src/sink/native.rs
+++ b/src/sink/native.rs
@@ -115,6 +115,10 @@ impl Sink<NativeConfig> for Native {
         let mut points = Vec::with_capacity(1024);
         let mut lines = Vec::with_capacity(1024);
 
+        if self.buffer.is_empty() {
+            return;
+        }
+
         for ev in self.buffer.drain(..) {
             match ev {
                 metric::Event::Telemetry(mut m) => {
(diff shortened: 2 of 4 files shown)


blt:
@blakebarnett / @tcng  FYI

PR #392, opened 2018-01-15
Fixes Nonblocking TcpStream Operations.

pulltab:
With the introduction of mio came nonblocking socket operations.  This
resulted in a regression in Avro, Native sources which read length-based
prefixed under the assumption of blocking IO.

This changeset introduces `BufferedPayload` as an interface for
nonblocking TCP sources which accept length-prefixed binary payloads.
`BufferedPayload` captures and buffers stream content, returning whole
payloads, one-at-a-time (when possible), when `read` is invoked.

If there is not enough data present to return a whole payload,
`PayloadErr::WouldBlock` is returned to the user.


src/source/avro.rs
```diff
@@ -134,47 +127,65 @@ impl TCPStreamHandler for AvroStreamHandler {
...
+                                match reader.read() {
+                                    Ok(payload) => {
+                                        let handle_res = self.handle_avro_payload(chans.clone(),tags, payload);
+                                        if handle_res.is_err() {
+                                            streaming = false;
+                                            break;
+                                        }
+
+                                        trace!("Avro payloads processed successfully.");
+                                            AVRO_PAYLOAD_SUCCESS_SUM
```
dparton: Extra indentation here

src/source/avro.rs
```diff
@@ -134,47 +127,65 @@ impl TCPStreamHandler for AvroStreamHandler {
...
-                                    stream
-                                        .write_all(resp.get_ref())
-                                        .expect("Failed to write response!");
+                        _stream_token => {
+                            while streaming {
+                                match reader.read() {
+                                    Ok(payload) => {
+                                        let handle_res = self.handle_avro_payload(chans.clone(),tags, payload);
+                                        if handle_res.is_err() {
+                                            streaming = false;
```
dparton: This probably calls for new internal telemetry

src/source/avro.rs
```diff
@@ -134,47 +127,65 @@ impl TCPStreamHandler for AvroStreamHandler {
...
+                                            resp.write_u64::<BigEndian>(id)
+                                                    .expect("Failed to write response id!");
+                                            stream
+                                                .write_all(resp.get_ref())
+                                                .expect("Failed to write response!");
+                                        }
+                                    }
+
+                                    Err(PayloadErr::WouldBlock) => {
+                                        //Not enough data on the wire.  Try again later.
```
dparton: Worth tracking this with internal telemetry?

src/source/native.rs
```diff
@@ -70,42 +69,59 @@ impl TCPStreamHandler for NativeStreamHandler {
...
-                            let rchans = chans.clone();
-                            self.handle_stream_payload(rchans, tags, &mut reader);
+                            while streaming {
+                                match reader.read() {
+                                    Ok(mut raw) => {
+                                        let rchans = chans.clone();
+                                        self.handle_stream_payload(rchans, tags, &mut raw);
+                                    }
+                                    Err(PayloadErr::WouldBlock) => {
+                                        //Not enough data yet.  Try again.
```
dparton: Telemetry?

src/source/native.rs
```diff
@@ -70,42 +69,59 @@ impl TCPStreamHandler for NativeStreamHandler {
...
+                                        break;
+                                    }
+                                    Err(PayloadErr::IO(ref e)) if e.kind() == ErrorKind::UnexpectedEof => {
+                                        //Client went away.  Shut it down (gracefully).
```
dparton: Telemetry?

src/source/native.rs
```diff
@@ -70,42 +69,59 @@ impl TCPStreamHandler for NativeStreamHandler {
...
+                                        trace!("TCP stream closed.");
+                                        streaming = false;
+                                        break;
+                                    }
+                                    Err(e) => {
+                                        error!("Failed to process native payload! {:?}", e);
```
dparton: Telemetry?

src/source/native.rs
```diff
@@ -70,42 +69,59 @@ impl TCPStreamHandler for NativeStreamHandler {
...
                     match event.token() {
                         constants::SYSTEM => return,
                         _stream_token => {
-                            let rchans = chans.clone();
-                            self.handle_stream_payload(rchans, tags, &mut reader);
+                            while streaming {
+                                match reader.read() {
+                                    Ok(mut raw) => {
+                                        let rchans = chans.clone();
+                                        self.handle_stream_payload(rchans, tags, &mut raw);
```
dparton: This has error conditions that don't seem to be returned / modeled. The Avro source is a little more sophisticated in this regard. Should these native source flows also handle errors instead of silently dropping them?

src/source/nonblocking.rs
```diff
@@ -0,0 +1,119 @@
...
+    remaining_bytes: Option<usize>,
+
+    /// Inner buffer where bytes from the underlying
+    /// stream are staged.
+    buffer: io::BufReader<mio::net::TcpStream>,
+}
+
+impl BufferedPayload {
+
+    /// Constructs a a new BufferedPayload
```
dparton: Comment typo

src/source/nonblocking.rs
```diff
@@ -0,0 +1,119 @@
...
+        // Are we actively reading a payload already?
+        if self.remaining_bytes.is_none() {
+            self.read_length()?
+        }
+
+        // At this point we can assume that we have successfully
+        // read the length off the wire.
+        let remaining_bytes = self.remaining_bytes.clone().unwrap();
+        let mut buf = Vec::new();
+        buf.resize(remaining_bytes, 0);
```
dparton: These 2 lines can be reduced to `let mut buf = vec![0u8; remaining_bytes];`

src/source/nonblocking.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// is constructed and returned.
+    ///
+    /// On non-blocking streams, it is up to the user to call
+    /// this method repeatedly until PayloadErr::WouldBlock
+    /// is returned.
+    pub fn read(&mut self) -> Result<Vec<u8>, PayloadErr>
+    {
+        // Are we actively reading a payload already?
+        if self.remaining_bytes.is_none() {
+            self.read_length()?
```
dparton: nitpick: Expression result isn't used, so this should technically have a semi-colon.

src/source/nonblocking.rs
```diff
@@ -0,0 +1,119 @@
...
+impl From<io::Error> for PayloadErr {
+    fn from(e: io::Error) -> PayloadErr {
+        match e.kind() {
+            io::ErrorKind::WouldBlock => {
+                PayloadErr::WouldBlock
+            }
+            _ => {
+                PayloadErr::IO(e)
+            }
+        }
```
dparton:
```
if e.kind() == io::ErrorKind::WouldBlock {
  PayloadErr::WouldBlock
} else {
  PayloadErr::IO(e)
}
```

is shorter

src/source/avro.rs
pulltab: Naw, I don't think so.

Par for the nonblocking stream course.

src/source/avro.rs
pulltab: I will run a cargo fmt.

src/source/native.rs
pulltab: Not for WouldBlock.

Expected to be a common occurrence.

src/source/native.rs
pulltab: None existed before.

Perhaps we pay down that tech debt in a separate pr?

src/source/native.rs
pulltab: None existed before.

Perhaps we pay down that tech debt in a separate pr?

src/source/nonblocking.rs
pulltab: Fixed.

src/source/nonblocking.rs
pulltab: Fixed.

src/source/nonblocking.rs
pulltab: Fixed

src/source/mod.rs
```diff
@@ -24,6 +25,7 @@ pub use self::file::{FileServer, FileServerConfig};
 pub use self::flush::{FlushTimer, FlushTimerConfig};
 pub use self::graphite::{Graphite, GraphiteConfig};
 pub use self::internal::{report_full_telemetry, Internal, InternalConfig};
+pub use self::nonblocking::{BufferedPayload, PayloadErr};
```
blt: Maybe `use self` rather than `pub use self`? Seems like these ought to be private to the source module, no?

src/source/native.rs
blt: It's expected that clients will come and go but I'm not opposed to telemetry around this.

src/source/nonblocking.rs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        // Are we actively reading a payload already?
+        if self.remaining_bytes.is_none() {
+            self.read_length()?;
+        }
+
+        // At this point we can assume that we have successfully
+        // read the length off the wire.
+        let remaining_bytes = self.remaining_bytes.clone().unwrap();
+        let mut buf = vec![0u8; remaining_bytes];
```
blt: Seems to me we could pass the buf in via read's arguments, save on allocating a vector for each read.

src/source/native.rs
blt: Agreed on doing this in a different PR. Worth an issue.

src/source/native.rs
```diff
@@ -70,42 +69,59 @@ impl TCPStreamHandler for NativeStreamHandler {
...
-                },
-            }
-        }
-    }
+                }, // events processing
+            } // poll
+        } // while connected
+
+        // On some systems shutting down an already closed connection (client or
+        // otherwise) results in an Err.  See -https://doc.rust-lang.org/beta/std/net/struct.TcpStream.html#platform-specific-behavior
```
blt: TIL

src/source/mod.rs
pulltab: Certainly.  Fixed.

src/source/nonblocking.rs
pulltab: The caller doesn't know the size of the payload beforehand.

src/source/native.rs
pulltab: I will make a ticket to beef up native's error logging + internal telemetry.

src/source/avro.rs
pulltab: Teased apart AVRO_PAYLOAD_PARSE_FAILURE_SUM from AVRO_PAYLOAD_IO_FAILURE_SUM.

src/source/nonblocking.rs
blt: Ah, got it. My thought won't work then.

PR #398, opened 2018-01-19
Removes Undesirable Panics From statsd/tcp Sources.

pulltab:
In order to support graceful shutdown, we require sources to be durable
in the presence of errors.  This change alters the following failure
scenarios:

* A tcp source fails to spawn a stream handler thread - instead of
panicking we now remove listener interfaces.  Effectively shutting down
the listener, reducing it to listening only for System events.

* A udp source fails to read from its socket - instead of panicking when a udp
socket operation returned an unsupported error, we now remove that
interface from the mio poller.  Allowing the source to still respond to
System events.

Note - file_server contains many `except`s and has not been addressed in
this PR.  I feel a specific refactoring is in order for file_server.

Lastly, sources continue to panic should polling for mio events ever
fails.  This is justified as without an ability to poll for mio events
the system cannot fulfill its mission.

### DIFF
--- a/src/source/statsd.rs
+++ b/src/source/statsd.rs
@@ -172,7 +172,7 @@ impl source::Source<StatsdConfig> for Statsd {
         poller: mio::Poll,
     ) -> () {
         for (idx, socket) in self.conns.iter() {
-            match poller
+            if let Err(e) = poller
                 .register(
                     socket,
                     mio::Token::from(idx),
@@ -180,10 +180,7 @@ impl source::Source<StatsdConfig> for Statsd {
                     mio::PollOpt::edge(),
                 )
             {
-                Ok(_) => {}
-                Err(e) => {
-                    error!("Failed to register {:?} - {:?}!", socket, e);
-                }
+                error!("Failed to register {:?} - {:?}!", socket, e);
             }
         }
 
@@ -200,17 +197,15 @@ impl source::Source<StatsdConfig> for Statsd {
 
                         token => {
                             let mut socket = &self.conns[token];
-                            match self.handle_datagrams(
-                                &mut chans,
-                                tags,
-                                &socket,
-                                &mut buf,
-                            ) {
-                                Ok(_) => {}
-                                Err(_fatal) => {
-                                    error!("Deregistering {:?} due to unrecoverable error!", *socket);
-                                    let _ = poller.deregister(socket);
-                                }
+                            if let Err(_e) = self
+                                .handle_datagrams(
+                                    &mut chans,
+                                    tags,
+                                    &socket,
+                                    &mut buf,
+                                )
+                            {
+                                error!("Deregistering {:?} due to unrecoverable error!", *socket);
                             }
                         }
                     }
--- a/src/source/statsd.rs
+++ b/src/source/statsd.rs
@@ -172,14 +172,19 @@ impl source::Source<StatsdConfig> for Statsd {
         poller: mio::Poll,
     ) -> () {
         for (idx, socket) in self.conns.iter() {
-            poller
+            match poller
                 .register(
                     socket,
                     mio::Token::from(idx),
                     mio::Ready::readable(),
                     mio::PollOpt::edge(),
                 )
-                .unwrap();
+            {
+                Ok(_) => {}
+                Err(e) => {
+                    error!("Failed to register {:?} - {:?}!", socket, e);
+                }
+            }
         }
 
         let mut buf = vec![0; 16_250];
--- a/src/source/statsd.rs
+++ b/src/source/statsd.rs
@@ -88,15 +88,19 @@ impl Default for StatsdConfig {
     }
 }
 
+enum StatsdHandlerErr {
+    Fatal,
+}
+
+
 impl Statsd {
     fn handle_datagrams(
-        &mut self,
+        & self,
         mut chans: &mut util::Channel,
         tags: &sync::Arc<metric::TagMap>,
-        token: mio::Token,
+        socket: &mio::net::UdpSocket,
         mut buf: &mut Vec<u8>,
-    ) {
-        let socket = &self.conns[token];
+    ) -> Result<(), StatsdHandlerErr>{
         let mut metrics = Vec::new();
         let basic_metric = sync::Arc::new(Some(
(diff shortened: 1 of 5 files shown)


src/source/statsd.rs
```diff
@@ -164,20 +166,25 @@ impl source::Source<StatsdConfig> for Statsd {
...
+            match poller
                 .register(
                     socket,
                     mio::Token::from(idx),
                     mio::Ready::readable(),
                     mio::PollOpt::edge(),
                 )
-                .unwrap();
+            {
+                Ok(_) => {}
```
dparton: Since we're not doing anything with the Ok arm, you can probably just use `if let Err(e) == poller.register... { error!....}`

src/source/statsd.rs
```diff
@@ -192,7 +199,19 @@ impl source::Source<StatsdConfig> for Statsd {
...
-                            self.handle_datagrams(&mut chans, tags, token, &mut buf);
+                            let mut socket = &self.conns[token];
+                            match self.handle_datagrams(
+                                &mut chans,
+                                tags,
+                                &socket,
+                                &mut buf,
+                            ) {
+                                Ok(_) => {}
+                                Err(_fatal) => {
```
dparton: Same here regarding `if let Err(e) == ...`

src/source/tcp.rs
```diff
@@ -106,14 +106,19 @@ where
         poller: mio::Poll,
     ) -> () {
         for (idx, listener) in self.listeners.iter() {
-            poller
+            match poller
```
dparton: Same here regarding `if let Err(e) ==`

src/source/tcp.rs
```diff
@@ -134,44 +139,54 @@ where
...
     fn spawn_stream_handlers(
-        &mut self,
+        &self,
         chans: util::Channel,
         tags: &sync::Arc<metric::TagMap>,
-        listener_token: mio::Token,
-    ) -> () {
-        let listener = &(self.listeners[listener_token]);
+        listener: &mio::net::TcpListener,
+    ) -> Result<Vec<thread::ThreadHandle>, std::io::Error> {
```
dparton: This is conveniently aliased `std::io::Result<Vec<thread::ThreadHandler>>`

pulltab: If it is all the same, I prefer as little magic as possible.

src/source/statsd.rs
pulltab: Fixed.

src/source/statsd.rs
pulltab: Fixed.

src/source/statsd.rs
blt: Endorse.

src/source/tcp.rs
```diff
@@ -134,44 +139,54 @@ where
...
+                            match self.spawn_stream_handlers(
+                                chans.clone(),
+                                tags,
+                                listener,
+                            ) {
+                                Ok(new_handlers) => {
+                                    self.handlers.extend(new_handlers);
+                                }
+                                Err(e) => {
+                                    error!("Failed to spawn stream handlers! {:?}", e);
```
blt: Under what conditions would we fail to spawn a stream handler? Far-side is partitioned away, too many files open. Anything else? 

Trying ultimately to figure out if this is a full-blown error logging condition – which people will ask about – or if it could be somewhat lower.

src/source/tcp.rs
```diff
@@ -134,44 +139,54 @@ where
...
+                        }
+                        listener_token => {
+                            let listener = &(self.listeners[listener_token]);
+                            match self.spawn_stream_handlers(
+                                chans.clone(),
+                                tags,
+                                listener,
+                            ) {
+                                Ok(new_handlers) => {
+                                    self.handlers.extend(new_handlers);
```
blt: Why does `spawn_stream_handlers` allocate a Vec for the new handlers if we're subsequently extending that vec onto `self.handlers`? I guess I'm missing something but it seems like we could avoid the allocation and keep tacking on to `self.handlers`?

src/source/tcp.rs
pulltab: All the cases when we fail to spawn a thread.  Maybe we are out of memory, fds, to name a couple.

src/source/tcp.rs
pulltab: There was an issue with immutable vs mutable borrows resulting from - https://github.com/postmates/cernan/pull/398/files/e04f650864a0da3a7fded0f7eeb13a09bd1ece63#diff-c657fbced7b730476f584243df38b7baR153.  That is what prompted me to have this function return a Vec.

I can pass in a ref to self.handlers instead.  Is there another way to resolve the conflict?

src/source/tcp.rs
dparton: `ConnectionAborted` is an error kind that should be ignored actually. That means a connection arrived in the listen queue but was subsequently closed prior to accept being called.

src/source/tcp.rs
blt: Hmm, let me fiddle some. Mind if I push a commit here? 

FWIW there's changing coming down the pipe in the borrow checker that will make these cases much less common.

src/source/tcp.rs
```diff
@@ -184,16 +196,17 @@ where
...
-                    self.handlers.push(new_stream);
+                    handlers.push(new_stream);
                 }
 
                 Err(e) => match e.kind() {
                     std::io::ErrorKind::WouldBlock => {
                         break;
                     }
-                    _ => unimplemented!(),
+                    _ => return Err(e),
```
dparton: `ConnectionAborted` is an error kind that should be ignored (by continuing) actually. That means a connection arrived in the listen queue but was subsequently closed prior to accept being called.

pulltab: Fixed.

dparton: Similarly, I believe we should `continue` with `Interrupted` and `TimedOut`

pulltab: Added all of the above.

dparton:
Lgtm. Lgty @blt?

PR #401, opened 2018-01-20
Improves Durability of Kinesis Sink.

pulltab:
Improvements are as follows:

* We no longer attempt to send a drained buffer via put_records.

* We now correctly detect when no publication failures occur in a batch.

* The kineis sink no longer panics, instead we generate a new client
when fatal exceptions are encountered.  This covers a number of cases
where AWS closes a connection prematurely.

* Buffering is now done optimally, directly into a PutRecordsInput struct.

### DIFF
--- a/src/sink/kinesis.rs
+++ b/src/sink/kinesis.rs
@@ -71,7 +71,7 @@ pub struct Kinesis {
     /// Number of bytes the current buffer represents.
     buffer_size: usize,
     /// Kinesis prepped event blobs.
-    buffer: Vec<PutRecordsRequestEntry>,
+    put_records_input: PutRecordsInput,
 }
 
 impl Sink<KinesisConfig> for Kinesis {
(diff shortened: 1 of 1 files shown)


src/sink/kinesis.rs
```diff
@@ -162,7 +167,7 @@ impl Kinesis {
 
         while !buffer.is_empty() {
             let put_records_input = PutRecordsInput {
-                records: self.buffer.clone(),
+                records: buffer.clone(),
```
blt: I wonder if you couldn't get higher throughput on this sink by relying only on `self.buffer`. Each time we go to `publish_buffer` we attempt to ship the whole thing, filter those records that were successfully sent etc etc. It would mean, I guess, potentially higher latency for any given buffer record, since you may not be able to keep the loop on 168. 

Might not be possible depending on the ownership semantics of PutRecordsInput?

pulltab: As I recall it wasn't possible due to PutRecordsInput.

pulltab: https://rusoto.github.io/rusoto/rusoto_kinesis/struct.PutRecordInput.html

It is just a Vec<u8>.  Though the structure is fully public.  I could use it as my buffer.

src/sink/kinesis.rs
```diff
@@ -69,7 +71,8 @@ pub struct Kinesis {
     /// Number of bytes the current buffer represents.
     buffer_size: usize,
     /// Kinesis prepped event blobs.
-    buffer: Vec<PutRecordsRequestEntry>,
+    put_records_input: PutRecordsInput,
+    //buffer: Vec<PutRecordsRequestEntry>,
```
blt: Intentionally left?

pulltab: Bah.  No.

PR #405, opened 2018-01-26
Kafka sink for raw events

dparton:
Adds a Kafka sink for publishing raw events, which should allow parity with Kinesis. This still needs some tests and I haven't run it locally against a kafka cluster. Looking for feedback on rust style, but also proper handling of the messages.

https://github.com/fede1024/rust-rdkafka
https://github.com/edenhill/librdkafka [particularly interesting: https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md ]
which errors are retry-able came from this table: http://kafka.apache.org/protocol.html#protocol_error_codes


src/config.rs
```diff
@@ -727,6 +731,91 @@ pub fn parse_config_file(buffer: &str, verbosity: u64) -> Args {
...
+                        res.config_path
+                    );
+                    continue;
+                }
+                res.brokers = brokers;
+
+                // Allow configuration of librdkafka producer with a sub-table.
+                // Sooo many options:
+                // https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
+                res.rdkafka_config = tbl.get("librdkafka").map(|x| {
```
blt: Hmm, this is tricky. On the one hand I'm concerned about exposing the underlying implementation through the config: if we choose to use a new library we've suddenly got compatibility issues. On the other hand, there _are_ a ton of configuration options here and it's nice not to have to work them into some abstract cernan state. 

I guess if we _did_ use another library at some point in the future the migration path would be support both, allow config passthrough in a like manner?

src/sink/kafka.rs
```diff
@@ -0,0 +1,214 @@
...
+            })
+            .collect();
+        let new_messages = retry_payload_and_keys
+            .iter()
+            .map(|maybe_retry| match *maybe_retry {
+                Some(ref message) => {
+                    let payload = message.payload();
+                    let key = message.key();
+                    if payload.is_some() && key.is_some() {
+                        Some(self.try_payload(payload.unwrap(), key.unwrap()))
```
blt: The unwrapping here makes me a touch nervous. Under what conditions would we hit the blow-up branch?

src/sink/kafka.rs
```diff
@@ -0,0 +1,214 @@
...
+
+        Kafka {
+            topic_name: config.topic_name.unwrap(),
+            producer: producer_config.create::<FutureProducer<_>>().unwrap(),
+            messages: Vec::new(),
+            flush_interval: config.flush_interval,
+        }
+    }
+
+    fn valve_state(&self) -> Valve {
```
blt: We may want to have some kind of valve. I forget what the overhead of a Future is but we're buffering them up into `self.messages`, exposing us to a potential unbounded memory buffer. 

Which, fancy trick, buffering Futures vs the payloads.

src/sink/kafka.rs
dparton: Should be no blow up here. The unwraps are guarded by the is_some checks in the if. TBH, I'm not sure under what conditions these things will come back as None. I should update the else branch to add to a counter so we can track if it ever happens with the internal telemetry.

src/sink/kafka.rs
dparton: I _think_ the flush function acts as a natural valve already to protect against a crazy amount of payloads already. In flush, each future is `wait`-ed upon, and by the time we return, only messages that failed and are now in-flight for retry will remain in the buffer.

src/config.rs
dparton: Yeah, I hear you. On the one hand, we don't necessarily want to co-opt all these settings, nor do we necessarily want the lib users to be stuck with only what we pull out. On the other hand, it's a PITA for those same users if we ever swap out the library. Perhaps if we ever switch libs, we would do what you say wrt to migration path and allow selection of the kafka lib via cargo.toml features?

src/sink/kafka.rs
blt: Ah ha, I missed the wait.

src/sink/kafka.rs
```diff
@@ -0,0 +1,214 @@
...
+        let key = format!("{:X}", order_by);
+        let future = self.try_payload(bytes.as_slice(), key.as_bytes());
+        self.messages.push(future);
+    }
+
+    fn flush(&mut self) {
+        let retry_payload_and_keys: Vec<Option<OwnedMessage>> = self.messages
+            .iter_mut()
+            .map(|future| {
+                let result = future.wait();
```
blt: I wonder if by waiting here we aren't shooting ourselves in the foot some. I _think_ what's going to happen is that rdkafka's IO loop will be blocked -- correct me if I'm wrong -- and we're going to see much worse throughput as a result. _If_ that is the case maybe what we ought to do every flush is troll through the futures in self.messages, poll them, if they're cleared out then remove them from self.messages? 

Would imply a need for a valve, still.

src/config.rs
blt: Sounds like a plan to me.

src/sink/kafka.rs
dparton: I'm happy to change this logic around if it ends up having crappy performance. First, I want to see (via a benchmark) how well it does perform in this simpler state though. 

As for the IO loop blocking, I don't think thats's the case. The future producer wraps a threaded producer and starts a thread that is constantly polling for the future's to complete:

Here's where the FutureProducer creates and wraps a ThreadedProducer:
https://github.com/fede1024/rust-rdkafka/blob/master/src/producer/future_producer.rs#L84

Here's where the ThreadedProducer is created and the thread started.
https://github.com/fede1024/rust-rdkafka/blob/master/src/producer/base_producer.rs#L244

The start loop is a few lines below there.

blt: Cool. That works back my concern then.

src/sink/kafka.rs
pulltab: +1 on the additional counter.

src/sink/kafka.rs
```diff
@@ -0,0 +1,214 @@
...
+
+                    _ => {
+                        error!("Failed in send to kafka broker: {:?}", result);
+                        KAFKA_PUBLISH_FAILURE_SUM.fetch_add(1, Ordering::Relaxed);
+                        None
+                    }
+                }
+            })
+            .collect();
+        let new_messages = retry_payload_and_keys
```
pulltab: Your rust-fu may be stronger than my own.

Where do we retry events that have failed republishing multiple times?

src/sink/kafka.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// batching internally.
+    fn deliver_raw(
+        &mut self,
+        order_by: u64,
+        _encoding: metric::Encoding,
+        bytes: Vec<u8>,
+    ) {
+        let key = format!("{:X}", order_by);
+        let future = self.try_payload(bytes.as_slice(), key.as_bytes());
+        self.messages.push(future);
```
pulltab: Why don't we bound on the size of messages here?

What saves us from blowing up a system by sending N events in a flush interval where N is 1 more event than the system can handle?

src/sink/kafka.rs
dparton: Latest commit adds valve control based on the number of bytes that are currently in flight

src/sink/kafka.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// Canonical name for the given Kafka sink.
+    pub config_path: Option<String>,
+    /// Kafka topic to publish to.
+    pub topic_name: Option<String>,
+    /// Kafka brokers. This is a comma-separated list of host or host:port.
+    pub brokers: Option<String>,
+    /// Underlying librdkafka configuration.
+    pub rdkafka_config: Option<HashMap<String, String>>,
+    /// Maximum number of bytes that can be in-flight. Once we go over this, the
+    /// valve closes. Default = DoubleEndedIteratorMb
```
dparton: Welp, that's an interesting "Autocorrect" -- this should read 10Mb.

src/sink/kafka.rs
dparton: There is now configurable valve control through `max_message_bytes` -- once the valve closes, deliver_raw stops getting called. So, we're now bounding (just not based on message count).

src/sink/kafka.rs
blt: Oh, interesting choice! Elsewhere in cernan the 'payload' is the standard metric, which was chosen vs. byte count arbitrarily. I guess if someone is willing to lose 100mb but has very tiny messages they may lose some, but an Ops person -- on the flip side -- will be able to guarantee the memory constraint. 

Hmm, makes me wonder now if we oughtn't to allow byte restrictions elsewhere, too.

src/sink/kafka.rs
```diff
@@ -0,0 +1,244 @@
...
+                                | RDKafkaError::NotLeaderForPartition
+                                | RDKafkaError::RequestTimedOut
+                                | RDKafkaError::NetworkException
+                                | RDKafkaError::GroupLoadInProgress
+                                | RDKafkaError::GroupCoordinatorNotAvailable
+                                | RDKafkaError::NotCoordinatorForGroup
+                                | RDKafkaError::NotEnoughReplicas
+                                | RDKafkaError::NotEnoughReplicasAfterAppend
+                                | RDKafkaError::NotController => {
+                                    KAFKA_PUBLISH_RETRY_SUM
```
blt: You should kick out a log message detailing the error, possibly even telemetry. TBH it was a real pain to kick out telemetry for Elasticsearch error variants but it's come in handy in practice.

blt: After discussing this in an IM, it turns out there are a _ton_ of possible options here. I am pro doing a sum-up telemetry plus logs for added detail.

src/sink/kafka.rs
```diff
@@ -0,0 +1,715 @@
...
+    fn get_retry(&self) -> usize {
+        0
+    }
+    fn get_retry_failed(&self) -> usize {
+        0
+    }
+}
+
+struct DefaultStatsCollector;
+impl StatsCollector for DefaultStatsCollector {
```
blt: This is a good idea.

PR #409, opened 2018-01-31
Don't double-count internal telemetry

blt:
We goofed when fiddling with the way we represent internal telemetry.
In the change we did three things:

  * change the aggregation of telemetry
  * remove zero-ing of the telemetry
  * issue zero points

The real problem comes when we issue the new aggregation, summing,
down to a sink that also performs aggregation. We're duplicating
values, potentially many times. Also, by issuing zero points we
increase aggregation load on the sink service for no real reason.

This commit takes us back to zero-ing and relies on the sink
aggregation to do the right thing. We also only emit a point if
it is non-zero.


blt:
@dparton this is what I was looking into. Figured out I wasn't taking crazy pills, with @benley's help.

PR #410, opened 2018-02-01
Attempt to get kcov working again by checking that files are executable

dparton:


blt:
@dparton this is cool as shit. Thank you for doing the leg work on this!

PR #415, opened 2018-02-15
Add new filter json_encode to convert loglines to json-encoded raw

ekimekim:
It will also optionally try to parse the log line as JSON and include
that at the top level.

This is needed for our structured logging so we can emit json
to disk and have cernan forward it to kafka.

### DIFF
--- a/src/filter/json_encode_filter.rs
+++ b/src/filter/json_encode_filter.rs
@@ -94,7 +94,7 @@ impl filter::Filter for JSONEncodeFilter {
                 // Data that is more likely to be correct (more specific to the source) overrides
                 // other data. So the parsed value is authoritative, followed by any fields we could
                 // parse by filters, then finally the metadata we were able to work out on our own.
-                let value = merge_objects(&[value, log.fields.clone().into(), metadata]);
+                let value = merge_objects(vec![value, log.fields.clone().into(), metadata]);
                 res.push(metric::Event::Raw {
                     order_by: random(),
                     encoding: metric::Encoding::JSON,
@@ -124,12 +124,12 @@ fn json_to_object(v: Value) -> Map<String, Value>  {
 /// Merge JSON objects, with values from earlier objects in the list overriding later ones.
 /// Note this is not a recursive merge - if the same key is in many objects, we simply take
 /// the value from the earliest one.
-fn merge_objects(objs: &[Map<String, Value>]) -> Map<String, Value> {
+fn merge_objects(objs: Vec<Map<String, Value>>) -> Map<String, Value> {
     let mut result = Map::new();
-    for obj in objs {
-        for key in obj.keys() {
-            if !result.contains_key(key) {
-                result.insert(key.clone(), obj[key].clone());
+    for obj in objs.into_iter() {
+        for (key, value) in obj.into_iter() {
+            if !result.contains_key(&key) {
+                result.insert(key, value);
             }
         }
     };
@@ -249,7 +249,7 @@ mod test {
     #[test]
     fn merged_objects_contain_all_keys() {
         fn inner(vecs: Vec<Vec<(String, String)>>) -> bool {
-            let result = merge_objects(vecs_to_objs(&vecs).as_slice());
+            let result = merge_objects(vecs_to_objs(&vecs));
             for obj in vecs {
                 for (k, _v) in obj {
                     if !result.contains_key(&k) {
@@ -266,7 +266,7 @@ mod test {
     fn merged_objects_takes_first_value() {
         fn inner(vecs: Vec<Vec<(String, String)>>) -> bool {
             let objs = vecs_to_objs(&vecs);
-            let result = merge_objects(objs.as_slice());
+            let result = merge_objects(objs.clone());
             for (key, result_value) in result {
                 match objs.iter().find(|obj| obj.contains_key(&key)) {
                     Some(obj) => if obj[&key] != result_value {
--- a/src/filter/json_encode_filter.rs
+++ b/src/filter/json_encode_filter.rs
@@ -1,0 +1,137 @@
+//! Convert LogLine events into Raw events encoded as JSON.
+//!
+//! This filter takes LogLines and encodes them into JSON, emitting the encoded
+//! event as a Raw event. This allows further filters or sinks to operate on the
+//! JSON without needing to understand a LogLine event in particular.
+//! If the LogLine value is a valid JSON object and parse_line config option is true,
+//! then the JSON will be merged with LogLine metadata. Otherwise, the original line
+//! will be included simply as a string.
+
+use chrono::DateTime;
+use chrono::naive::NaiveDateTime;
+use chrono::offset::Utc;
+use filter;
+use metric;
+use metric::TagMap;
+use rand::random;
+use serde_json;
+use serde_json::Value;
+use serde_json::map::Map;
+use std::iter::FromIterator;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+lazy_static! {
+    /// Total number of logline processed
+    pub static ref JSON_ENCODE_LOG_PROCESSED: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
+    /// Total number of logline with JSON value successfully parsed
+    pub static ref JSON_ENCODE_LOG_PARSED: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
+}
+
+impl From<TagMap> for Map<String, Value> {
+    fn from(tagmap: TagMap) -> Self {
+        Map::from_iter(
+            tagmap.into_iter().map(|(k, v)| (k.clone(), v.clone().into()))
+        )
+    }
+}
+
+/// Convert LogLine events into Raw events encoded as JSON.
+///
+/// This filter takes LogLines and encodes them into JSON, emitting the encoded
+/// event as a Raw event. This allows further filters or sinks to operate on the
+/// JSON without needing to understand a LogLine event in particular.
+/// If the LogLine value is a valid JSON object and parse_line config option is true,
+/// then the JSON will be merged with LogLine metadata. Otherwise, the original line
+/// will be included simply as a string.
+pub struct JSONEncodeFilter {
+    parse_line: bool,
+}
+
+/// Configuration for `JSONEncodeFilter`
+#[derive(Clone, Debug)]
+pub struct JSONEncodeFilterConfig {
+    /// The filter's unique name in the routing topology.
+    pub config_path: Option<String>,
+    /// The forwards along which the filter will emit its `metric::Event`s.
+    pub forwards: Vec<String>,
+    /// Whether the filter should attempt to parse LogLine values that are valid JSON objects.
+    pub parse_line: bool,
+}
+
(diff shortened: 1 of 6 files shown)


src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,139 @@
...
+            },
+            metric::Event::TimerFlush(f) => {
+                res.push(metric::Event::TimerFlush(f));
+            }
+            raw @ metric::Event::Raw { .. } => {
+                res.push(raw);
+            }
+            metric::Event::Shutdown => {
+                res.push(metric::Event::Shutdown);
+            }
```
dparton: It looks to me like these last 3 match arms are all doing the same thing, so this could be simplified to:
```
event => { 
    res.push(event); 
}
```

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,139 @@
...
+use serde_json::map::Map;
+use std::iter::FromIterator;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+lazy_static! {
+    /// Total number of logline processed
+    pub static ref JSON_ENCODE_LOG_PROCESSED: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
+    /// Total number of logline with JSON value successfully parsed
+    pub static ref JSON_ENCODE_LOG_PARSED: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
```
dparton: We're counting these things but you haven't added any code to publish them in the internal source. Check out `src/source/internal.rs`

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,139 @@
...
+                // Otherwise get an object containing the original line.
+                let value = (if self.parse_line { Some(()) } else { None })
+                    .and_then(|_| serde_json::from_str::<Value>(&log.value).ok())
+                    .and_then(|v| if let Value::Object(obj) = v { Some(obj) } else { None })
+                    .map(|v| { JSON_ENCODE_LOG_PARSED.fetch_add(1, Ordering::Relaxed); v })
+                    .unwrap_or_else(||
+                        if let Value::Object(obj) = json!({"message": log.value.clone()}) {
+                            obj
+                        } else {
+                            panic!()
```
dparton: Should this be `unreachable!()`?

src/filter/json_encode_filter.rs
ekimekim: ah, right. I just copied this part from DelayFilter.

ekimekim:
PTAL

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl filter::Filter for JSONEncodeFilter {
+    fn process(
+        &mut self,
+        event: metric::Event,
+        res: &mut Vec<metric::Event>,
+    ) -> Result<(), filter::FilterError> {
+        match event {
+            metric::Event::Telemetry(_) => {
+                // Drop telemetry
```
blt: Rather than drop we should probably forward Telemetry through.

ekimekim: fair enough. wasn't sure of the convention for filters that don't apply to a particular type.

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,133 @@
...
+            event => {
+                res.push(event);
+            }
+        }
+        Ok(())
+    }
+}
+
+/// Merge JSON objects, with values from earlier objects in the list overriding later ones.
+/// Note this is not a recursive merge - if the same key is in many objects, we simply take
```
blt: This type of property is ripe for a quickcheck test.

src/filter/mod.rs
```diff
@@ -13,9 +13,11 @@ use util;
 mod programmable_filter;
 pub mod delay_filter;
 mod flush_boundary_filter;
+mod json_encode_filter;
 
 pub use self::delay_filter::{DelayFilter, DelayFilterConfig};
 pub use self::flush_boundary_filter::{FlushBoundaryFilter, FlushBoundaryFilterConfig};
+pub use self::json_encode_filter::{JSONEncodeFilter, JSONEncodeFilterConfig};
```
blt: These are now exposed but you'll need to add them to `src/config.rs` to hook up configuration.

src/filter/json_encode_filter.rs
blt: We don't have a concrete policy around this and we should. I lean toward "pass forward unless otherwise documented".

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,133 @@
...
+                    .map(|v| { JSON_ENCODE_LOG_PARSED.fetch_add(1, Ordering::Relaxed); v })
+                    .unwrap_or_else(||
+                        if let Value::Object(obj) = json!({"message": log.value.clone()}) {
+                            obj
+                        } else {
+                            unreachable!()
+                        }
+                    );
+                // Combine our various sources of data. Metadata overrides parsed values overrides user-specified.
+                let value = merge_objects(&[&metadata, &log.fields.clone().into(), &value]);
```
blt: I think as a user I'd find this confusing. If my json blob had "time" in it then it'd never come through on the other side. This'd be especially tricky if it _mostly_ did roll through the other side. 

What's stopping us from _adding_ to parsed structures? So, make a `__cernan_metadata` or whatever name field in the parsed blob that contained the contents of metadata? The users' data comes through the other side unchanged, we get our data too. 
Maybe

blt:
Agreed RE testing. Also, bonus points for adding a benchmark for `JSONEncodeFilter::process`.

src/filter/json_encode_filter.rs
ekimekim: The primary use-case of this is for searching in logs. In that context the key names need to be short, simple and memorable.
There's probably a good argument for allowing the user to override cernan's metadata rather than vice versa - eg. they probably know the true timestamp of the event better than we do (our view of time is just when we read it).
In fact, yeah I think that ordering makes more sense. The user knows best, anything we can clean from filters (fields) is second best, and our own metadata is third.

ekimekim:
PTAL

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    // quickcheck and serde_json::map::Map aren't compatible, so we ask quickcheck for
+    // many Vec<(String, String)>s that we turn into maps.
+    fn vecs_to_objs(vecs: &Vec<Vec<(String, String)>>) -> Vec<Map<String, Value>> {
+        iter(vecs).map(|vec|
+            iter(vec).map(|(k, v)| (k.clone(), v.clone().into()))
+        ).into()
+    }
+
+    quickcheck! {
```
blt: For consistency with the rest of the codebase could you flip this over to use the builder style pattern? A good example of what I mean: https://github.com/postmates/cernan/blob/master/src/buckets.rs#L383

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,268 @@
...
+                }
+            }
+            true
+        }
+
+        fn merged_objects_takes_first_value(vecs: Vec<Vec<(String, String)>>) -> bool {
+            let objs = vecs_to_objs(&vecs);
+            let result = merge_objects(&objs);
+            for (key, result_value) in result {
+                match iter(objs).find(|obj| obj.contains_key(key)) {
```
blt: This `iter(objs)` is going to cause a build error I do believe.

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,268 @@
...
+                "time": "2000-01-01T00:00:00",
+                "tags": {},
+            })
+        );
+    }
+
+    // quickcheck and serde_json::map::Map aren't compatible, so we ask quickcheck for
+    // many Vec<(String, String)>s that we turn into maps.
+    fn vecs_to_objs(vecs: &Vec<Vec<(String, String)>>) -> Vec<Map<String, Value>> {
+        iter(vecs).map(|vec|
```
blt: `vecs.iter()`?

src/filter/json_encode_filter.rs
ekimekim: yep, there's a bunch of build issues i'm just dealing with now, i was trying to parallelise fixing those with higher-level feedback :P

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,268 @@
...
+            iter(vec).map(|(k, v)| (k.clone(), v.clone().into()))
```
blt: It doesn't matter in-test, but all the cloning here could be avoided if `vecs_to_objs` consumed `vecs`. Right now you're passing by reference and then cloning data from inside the clone.

src/filter/json_encode_filter.rs
```diff
@@ -0,0 +1,268 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+/// Merge JSON objects, with values from earlier objects in the list overriding later ones.
+/// Note this is not a recursive merge - if the same key is in many objects, we simply take
+/// the value from the earliest one.
+fn merge_objects(objs: &[&Map<String, Value>]) -> Map<String, Value> {
```
blt: You're passing `objs` into this function but then the passed values are never re-used. Consider moving `objs` into this function so you can avoid cloning on line 129.

src/filter/json_encode_filter.rs
ekimekim: yeah. I was able to figure that out, but couldn't figure out how to make a consuming iterator such that `k` and `v` would be `String` instead of `&String`.

src/filter/json_encode_filter.rs
blt: > The primary use-case of this is for searching in logs. In that context the key names need to be short, simple and memorable.

That's fair. Once it's merged in master it's not really about the originating use-case anymore and can be applied in surprising ways. That's the perch I'm sitting on when discussing stuff, anyhow. 

> There's probably a good argument for allowing the user to override cernan's metadata rather than vice versa - eg. they probably know the true timestamp of the event better than we do (our view of time is just when we read it).

That's possible to do with a downstream filter. Possible I'm missing some further elaboration you're imagining. 

> In fact, yeah I think that ordering makes more sense. The user knows best, anything we can clean from filters (fields) is second best, and our own metadata is third.

I guess I'm still a little worried about overwriting fields in a payload rather than offsetting them in some way. Adding fields, sure, that's okay but overwriting strikes me as something that could cause surprises, especially for some user that doesn't have logging in mind.

src/filter/json_encode_filter.rs
blt: Try `into_iter`.

ekimekim:
Added tests, including quickchecks for two properties of merge_object (that all keys in the source objects are present in the merged object, and that all keys in the merged object map to the first value found in the list of source objects).
Changed the ordering of the merge so it goes (parsed line, fields, metadata).

PTAL?

ekimekim:
rebased onto latest master. no conflicts.

ekimekim:
(skipped waiting for tests as there was zero diff between old and new version, just commit history edits)

Issue #406, opened 2018-01-27
Consider setting `panic=abort` for release builds


blt:
We've had repeat reports that errors in one part of cernan will leave it partially available: a thread'll abort but the upstream nodes will keep pushing into queues etc etc. 

If we set `panic=abort` then all of cernan will go offline but the startup harness will be able to choose to restart it.

blt:
@blakebarnett @pulltab @dparton thoughts?

tfheen:
I think just exiting so the init system can restart it makes a lot of sense, having it half-live has very limited value.

pulltab:
I believe this is an easy fix which will solve one class of failures.  The more complete solution is to be able to detect all situations where a worker thread is failing to make forward progress.  This will require application layer heartbeats and supervision of subordinate threads.

PR #416, opened 2018-02-16
Force release builds to abort on panic

blt:
This corrects #406. As @pulltab notes there's a cleaner way of
recovering from internal failures but we'll opt for now to let
the init system or what not catch us when we fall.


blt:
@tfheen @pulltab FYI

PR #417, opened 2018-02-17
Update hopper to 0.4

### LINKED ISSUE #412: Ingress Shedding
We'd like to support ingress shedding in cernan. This is distinct from our previous model in which all input data is considered to be delivered according to best effort and is never dropped. Ingress shedding will be _opt-in_ and configurable. 

The notion I have running now is, in addition to forwards, you configure the capacity of the forward link. If that capacity overruneth you lose your data. There will be telemetry around all of this to allow Ops to quantify the impact. Related to #408

blt:
This commit incorporates changes made to hopper to allow for ingress shedding in overload conditions. We expose these configuration parameters in cernan, allowing operations to choose to shed data rather than eat up disk space until a crash.

By default each hopper queue file will consume 1_048_576 * 100 * 100 bytes on disk, at a maximum.

This resolves #412.

This cannot be merged until https://github.com/postmates/hopper/pull/20 lands and hopper 0.4.0 is properly cut.

We've removed the reference-count encoding via serde. It forced a copy and interacted badly with the new hopper. In particular, serializing Rcs and shipping them through hopper causes a memory leak for reasons that are not clear to me. Serde shies you off from serializing Rcs and I can't say I much disagree.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -61,6 +61,13 @@ fn default_version() -> String {
 /// on `parse_args` in this module for more details.
 #[derive(Debug)]
 pub struct Args {
+    /// The maximum size -- in bytes -- that a hopper queue may hold in memory
+    /// before flushing to disk.
+    pub max_hopper_in_memory_bytes: usize,
+    /// The maximum number of queue files that hopper may hold on disk. The
+    /// maximum discu consumption of a single hopper queue will be
+    /// `max_hopper_queue_files * max_hopper_queue_bytes`.
+    pub max_hopper_queue_files: usize,
     /// The maximum size -- in bytes -- that a hopper queue may grow to before
     /// being cycled.
     pub max_hopper_queue_bytes: usize,
@@ -127,7 +134,9 @@ pub struct Args {
 impl Default for Args {
     fn default() -> Self {
         Args {
+            max_hopper_in_memory_bytes: 1_048_576,
             max_hopper_queue_bytes: 1_048_576 * 100,
+            max_hopper_queue_files: 100,
             data_directory: default_data_directory(),
             scripts_directory: default_scripts_directory(),
             flush_interval: 60,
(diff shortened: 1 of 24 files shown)


src/config.rs
```diff
@@ -61,6 +61,13 @@ fn default_version() -> String {
 /// on `parse_args` in this module for more details.
 #[derive(Debug)]
 pub struct Args {
+    /// The maximum size -- in bytes -- that a hopper queue may hold in memory
+    /// before flushing to disk.
+    pub max_hopper_in_memory_bytes: usize,
+    /// The maximum number of queue files that hopper may hold on disk. The
+    /// maximum discu consumption of a single hopper queue will be
```
dparton: Minor type in the comment here `discu`

dparton: Ahhh typoing typo. The irony is not lost on me.

src/filter/delay_filter.rs
```diff
@@ -59,26 +59,24 @@ impl filter::Filter for DelayFilter {
...
-                    res.push(metric::Event::new_log(log));
+                    res.push(metric::Event::Log(log));
                 } else {
                     DELAY_LOG_REJECT.fetch_add(1, Ordering::Relaxed);
                 }
-            },
-            metric::Event::TimerFlush(f) => {
-                res.push(metric::Event::TimerFlush(f));
+            }
+            timer @ metric::Event::TimerFlush { .. } => {
```
dparton: The next 3 match arms are all the same, so they can just be `metric => { res.push(metric); }`

src/sink/null.rs
```diff
@@ -35,11 +34,11 @@ impl Sink<NullConfig> for Null {
         Valve::Open
     }
 
-    fn deliver(&mut self, _: sync::Arc<Option<Telemetry>>) -> () {
+    fn deliver(&mut self, _: Telemetry) -> () {
```
dparton: This and `delivery_line` can be removed with the new default impls in the Sink trait.

PR #423, opened 2018-02-28
Emit debug, not error from Elasticsearch sink

blt:
We have a number of error notifs in the ES sink that are not
really errors. This is now corrected by flipping these things
to debug.

### DIFF
--- a/src/sink/elasticsearch.rs
+++ b/src/sink/elasticsearch.rs
@@ -228,13 +228,13 @@ impl Sink<ElasticsearchConfig> for Elasticsearch {
                                 ELASTIC_RECORDS_TOTAL_FAILED
                                     .fetch_add(1, Ordering::Relaxed);
                                 if let Some(cause) = item.cause() {
-                                    error!(
+                                    debug!(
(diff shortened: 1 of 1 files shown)


src/sink/elasticsearch.rs
```diff
@@ -228,13 +228,13 @@ impl Sink<ElasticsearchConfig> for Elasticsearch {
                                 ELASTIC_RECORDS_TOTAL_FAILED
                                     .fetch_add(1, Ordering::Relaxed);
                                 if let Some(cause) = item.cause() {
-                                    error!(
+                                    debug!(
```
pulltab: Why don't we want to know about these in normal production?

pulltab: I suppose the answer is that we already have internal telemetry so the logs are redundant.

blt: They're useful for debugging purposes but if you kick them out at an error level _and_ point cernan at itself you get a feedback loop. 

None of these are actionable, really. Just useful information to have for fiddling.

PR #427, opened 2018-03-01
Fix the lack of tags in ES sink, introduced in 0.8.13

blt:
This commit repairs a defect with 0.8.13/14 in which tags were not
being correctly distributed to ES and Firehose sinks, causing LogLines
traveling out through those sinks to go out tag-less.

LogLine now follows Telemetry and does not bundle tags by default. There's
a trait in here somewhere, TaggableEvent or some such. Should cut down
on the code duplication.

### DIFF
--- a/src/filter/programmable_filter.rs
+++ b/src/filter/programmable_filter.rs
@@ -170,14 +170,16 @@ impl<'a> Payload<'a> {
         let pyld = state.to_userdata(1) as *mut Payload;
         let idx = idx(state.to_integer(2), (*pyld).logs.len());
         match state.to_str(3).map(|k| k.to_owned()) {
-            Some(key) => match (*pyld).logs[idx].tags.get(&key) {
-                Some(v) => {
-                    state.push_string(v);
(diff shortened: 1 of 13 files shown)


src/filter/programmable_filter.rs
```diff
@@ -267,7 +269,7 @@ impl<'a> Payload<'a> {
         let idx = idx(state.to_integer(2), (*pyld).logs.len());
         match state.to_str(3).map(|k| k.to_owned()) {
             Some(key) => match state.to_str(4).map(|v| v.to_owned()) {
-                Some(val) => match (*pyld).logs[idx].tags.insert(key, val) {
+                Some(val) => match (*pyld).logs[idx].insert_tag(key, val) {
```
pulltab: Can you explain the semantic difference here?

blt:
@pulltab still need to toss info through to elasticsearch.

blt:
@pulltab confirmed by inspection that logs are making it through with their correct tags as of bae5a9f.

PR #430, opened 2018-03-05
Remove extra lazy_statics

blt:
Turns out our use of lazy_static is kind of anachronistic. There's
more that can be done in current stable with regard to static
initialization than there used to be. To that end, most of the
use of lazy_static is now gone and we rely on statically initializable
AtomicUsizes.

### DIFF
--- a/src/filter/delay_filter.rs
+++ b/src/filter/delay_filter.rs
@@ -7,20 +7,17 @@
 
 use filter;
 use metric;
-use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use time;
 
-lazy_static! {
-    /// Total number of telemetry rejected for age
-    pub static ref DELAY_TELEM_REJECT: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-    /// Total number of telemetry accepted for age
-    pub static ref DELAY_TELEM_ACCEPT: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-    /// Total number of logline rejected for age
-    pub static ref DELAY_LOG_REJECT: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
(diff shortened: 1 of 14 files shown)


src/filter/delay_filter.rs
```diff
@@ -7,20 +7,17 @@
...
-    pub static ref DELAY_TELEM_REJECT: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-    /// Total number of telemetry accepted for age
-    pub static ref DELAY_TELEM_ACCEPT: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-    /// Total number of logline rejected for age
-    pub static ref DELAY_LOG_REJECT: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-    /// Total number of logline accepted for age
-    pub static ref DELAY_LOG_ACCEPT: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
-}
+/// Total number of telemetry rejected for age
+pub static DELAY_TELEM_REJECT: AtomicUsize = AtomicUsize::new(0);
```
pulltab: What effect does removing ref here have?

src/time.rs
```diff
@@ -6,11 +6,10 @@
 
 use chrono::offset::Utc;
 use std::{thread, time};
-use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 lazy_static! {
```
pulltab: Why does this one remain?

src/filter/delay_filter.rs
blt: The ref is needed by the macro. Plain `static` doesn't need it.

src/time.rs
blt: Ah! For a thing to be static all the initialization code has to be callable at compile-time. Can't read the time at compilation time, so we gotta initialize lazily.

PR #431, opened 2018-03-06
Add error handling to calling lua functions

ekimekim:
Without this, errors in lua cause the C lua code to abort().

I found it was easier to refactor the repeated
process_metric/process_log/tick code than to make the change in three
places.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -26,17 +26,21 @@ pub enum FilterError {
     /// Specific to a ProgrammableFilter, means no function is available as
     /// called in the script
     NoSuchFunction(&'static str, metric::Event),
+    /// Specific to a ProgrammableFilter, means the lua code errored
+    LuaError(String, metric::Event),
 }
 
-fn name_in_fe(fe: &FilterError) -> &'static str {
-    match *fe {
-        FilterError::NoSuchFunction(n, _) => n,
+fn msg_in_fe(fe: &FilterError) -> &str {
+    match fe {
+        &FilterError::NoSuchFunction(n, _) => n,
+        &FilterError::LuaError(ref n, _) => n,
     }
 }
 
 fn event_in_fe(fe: FilterError) -> metric::Event {
     match fe {
         FilterError::NoSuchFunction(_, m) => m,
+        FilterError::LuaError(_, m) => m,
     }
 }
 
@@ -95,7 +99,7 @@ pub trait Filter {
                         Err(fe) => {
                             error!(
                                 "Failed to run filter with error: {:?}",
-                                name_in_fe(&fe)
+                                msg_in_fe(&fe)
                             );
                             let event = event_in_fe(fe);
                             util::send(&mut chans, event);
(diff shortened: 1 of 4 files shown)


blt:
Nice. I'll review this directly.

src/filter/mod.rs
```diff
@@ -26,17 +26,21 @@ pub enum FilterError {
...
     NoSuchFunction(&'static str, metric::Event),
+    /// Specific to a ProgrammableFilter, means the lua code errored
+    LuaError(String, metric::Event),
 }
 
-fn name_in_fe(fe: &FilterError) -> &'static str {
-    match *fe {
-        FilterError::NoSuchFunction(n, _) => n,
+fn msg_in_fe(fe: &FilterError) -> &str {
+    match fe {
```
blt: This could be `match *fe` and then you wouldn't need to match on references.

src/filter/mod.rs
```diff
@@ -26,17 +26,21 @@ pub enum FilterError {
...
+        &FilterError::NoSuchFunction(n, _) => n,
+        &FilterError::LuaError(ref n, _) => n,
     }
 }
 
 fn event_in_fe(fe: FilterError) -> metric::Event {
     match fe {
         FilterError::NoSuchFunction(_, m) => m,
```
blt: These arms could be combined, so 

```
FilterError::NoSuchFunction(_, m) | FilterError::LuaError(_, m) => m
```

Same with `msg_in_fe` above.

src/filter/mod.rs
ekimekim: That's how it used to work, but I hit issues with lifetime because LuaError contains a `String`, not a `&'static str`. Trying to return a borrow of the deconstructed `String` said that it didn't live past the end of this function.

ekimekim:
I don't understand why the codecov is failing - the new code is mostly the moving of old code, which is exercised just as much now as before. And even the new code paths are exercised by the new test.

src/filter/mod.rs
blt: Oh shoot, of course.

blt:
@ekimekim I think it has to do with every lua test being an integration test. We have to feed the codecov uploader the exact binary names. Hmm. No, but that should be taken care of in the `.travis.yml` already in `after_success`. 

@dparton do you have any ideas here?

blt:
@ekimekim I'm pro this PR, by the way. Tidy work. I'd like to have the codecov thing figured out before we merge up is all.

src/filter/mod.rs
ekimekim: compiler didn't like the `msg_in_fe` one:
```
  --> src/filter/mod.rs:35:73
   |
35 |         &FilterError::NoSuchFunction(n, _) | &FilterError::LuaError(ref n, _) => n,
   |                                      - first binding                    ^ bound in different ways

error[E0409]: variable `n` is bound in inconsistent ways within the same match arm
```
but done for event_in_fe.

blt:
Oh dang, so our integration tests don't get picked up by codecov on account of their naming scheme. Worth resolving but not as a part of this PR.

ekimekim:
merging despite codecov reject because the diff is covered well by integration tests, which codecov doesn't detect.

PR #433, opened 2018-03-06
Add a context that ignores "Receive failed: Disconnected" error logs

dparton:
Adds a context that overrides the default impls for emitting errors. Also, I ran rustfmt on the code, but it failed to format `internal.rs`.

### DIFF
--- a/src/filter/json_encode_filter.rs
+++ b/src/filter/json_encode_filter.rs
@@ -3,9 +3,10 @@
 //! This filter takes `LogLines` and encodes them into JSON, emitting the
 //! encoded event as a Raw event. This allows further filters or sinks to
 //! operate on the JSON without needing to understand a `LogLine` event in
-//! particular.  If the ``LogLine`` value is a valid JSON object and `parse_line`
-//! config option is true, then the JSON will be merged with `LogLine`
-//! metadata. Otherwise, the original line will be included simply as a string.
+//! particular.  If the `LogLine` value is a valid JSON object and
+//! `parse_line` config option is true, then the JSON will be merged with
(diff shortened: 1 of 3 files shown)


src/filter/json_encode_filter.rs
```diff
@@ -3,9 +3,10 @@
 //! This filter takes `LogLines` and encodes them into JSON, emitting the
 //! encoded event as a Raw event. This allows further filters or sinks to
 //! operate on the JSON without needing to understand a `LogLine` event in
-//! particular.  If the ``LogLine`` value is a valid JSON object and `parse_line`
-//! config option is true, then the JSON will be merged with `LogLine`
-//! metadata. Otherwise, the original line will be included simply as a string.
+//! particular.  If the ``LogLine`` value is a valid JSON object and
```
blt: '``LogLine``' should be '`LogLine`'.

99% sure that's my search-and-replace goof.

src/sink/kafka.rs
```diff
@@ -24,6 +24,60 @@ lazy_static! {
     pub static ref KAFKA_PUBLISH_RETRY_FAILURE_SUM: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
 }
 
+struct STFUContext;
```
blt: Would also have accepted `SssshhhContext`.

src/filter/json_encode_filter.rs
dparton: Eh?

dparton: Oh I see now when Editing your message. Change double backtick to single backtick.

blt: Heh, markdown is hard.

PR #446, opened 2018-06-27
Fix TCP Stream Handler State Leak

pulltab:
TCP stream handler state was never garbage collected on disconnect
resulting in memory exhaustion after sufficient TCP life cycles.

The following changes have been made to enable this garbage collection:

* source/tcp now creates a special registration and token which
corresponds to all stream handler events.

* src/thread has been extended with a new `spawn2` function which operates
as `spawn`, but takes an additional `mio::SetReadiness` which is flagged
readable after the given closure returns control.

* thread::Stoppable has been extended to include a new `ready` function
which returns true when joining the given thread is guaranteed not block.
`ready` always returns false on threads created with `spawn`.

### DIFF
--- a/src/source/tcp.rs
+++ b/src/source/tcp.rs
@@ -4,6 +4,7 @@ use metric;
 use mio;
 use source::Source;
 use std;
+use std::collections;
 use std::io::ErrorKind;
 use std::marker::PhantomData;
 use std::net::ToSocketAddrs;
@@ -43,6 +44,11 @@ pub trait TCPStreamHandler: 'static + Default + Clone + Sync + Send {
         Default::default()
     }
 
+    /// Invokes handle_stream, toggling readiness if the function returns control.
+    fn run(&mut self, chan: util::Channel, poll: &mio::Poll, stream: mio::net::TcpStream) -> () {
+        self.handle_stream(chan, poll, stream);
+    }
+
     /// Handler for a single HTTP request.
     fn handle_stream(&mut self, util::Channel, &mio::Poll, mio::net::TcpStream) -> ();
 }
@@ -50,7 +56,10 @@ pub trait TCPStreamHandler: 'static + Default + Clone + Sync + Send {
 /// State for a TCP backed source.
 pub struct TCP<H> {
     listeners: util::TokenSlab<mio::net::TcpListener>,
-    handlers: Vec<thread::ThreadHandle>,
+    stream_events: mio::Registration,
+    stream_events_token: mio::Token,
+    stream_events_readiness: mio::SetReadiness,
+    handlers: collections::HashMap<usize, thread::ThreadHandle>,
     phantom: PhantomData<H>,
 }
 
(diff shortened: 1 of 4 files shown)


src/source/tcp.rs
```diff
@@ -151,23 +194,28 @@ where
...
+                                    mio::Token(0),
+                                    mio::Ready::readable(),
+                                    mio::PollOpt::edge(),
+                                )
+                                .unwrap();
+
+                            let mut handler = H::new();
+                            handler.run(rchans, &poller, stream);
+                        });
+                    let handler_id = self.handlers.len();
```
dparton: Subtle bug here; this needs to be a monotonically increasing counter. 

For example:
Imagine you have 4 clients, then the first 2 disconnect. The next 2 connections will get the same value as the remaining 2.

src/source/tcp.rs
```diff
@@ -151,23 +194,28 @@ where
...
-                                mio::Ready::readable(),
-                                mio::PollOpt::edge(),
-                            )
-                            .unwrap();
-
-                        let mut handler = H::new();
-                        handler.handle_stream(rchans, &poller, stream);
-                    });
-                    self.handlers.push(new_stream);
+                    let stream_events_readiness = self.stream_events_readiness.clone();
```
dparton: Something seems off with this. We're cloning this thing here and passing it in, but we're cloning it again in spawn2... how is this meant to be used?

src/thread.rs
```diff
@@ -50,23 +54,106 @@ pub fn spawn<F>(f: F) -> ThreadHandle
...
+        self.thread_id += 1;
+        thread_id
+    }
+
+    /// Block on completion of all executing threads.
+    pub fn join(mut self) -> Vec<usize>
+    {
+        self.threads.drain().for_each(|(_, h)| h.join());
+        self.join_ready()
+    }
```
dparton: This function looks incorrect... it looks like shutdown does the correct thing. `self.threads.drain` will empty the container, but then `join_ready` will attempt to `let handle = self.threads.remove(&id).unwrap(); handle.join()` below.

pulltab: Good catch.  I will wrap the remove in a guard.

pulltab: Updated.

src/source/tcp.rs
```diff
@@ -43,14 +42,21 @@ pub trait TCPStreamHandler: 'static + Default + Clone + Sync + Send {
         Default::default()
     }
 
+    /// Invokes handle_stream, toggling readiness if the function returns control.
+    fn run(&mut self, chan: util::Channel, poll: &mio::Poll, stream: mio::net::TcpStream) -> () {
```
pulltab: This function needs to be cleaned up.  We can resume just calling handle_stream directly.

src/thread.rs
```diff
@@ -50,23 +54,107 @@ pub fn spawn<F>(f: F) -> ThreadHandle
...
+        while let Some(id) = joinable.pop() {
+            if let Some(handle) = self.threads.remove(&id) {
+                handle.join();
+            }
+            joined.push(id);
+        }
+        joined
+    }
+
+    /// Signal shutdown and block for their completion.
```
pulltab: This comment is awkward.

pulltab: Fixed.

src/source/tcp.rs
pulltab: Removed.

blt:
@pulltab looking at this now.

pulltab:
@blt Cool.  Happy to have a call about this and how it can fit into larger improvements.

PR #448, opened 2018-08-06
Adds synchronous acknowledge to the avro->kafka path

dparton:
This changes the avro source to wait for sink acknowledgement of messages that are marked as synchronous. For now, this means that at least one sink has acknowledged the message.

Highlights:
* Each tcp stream connected to an avro source generates a UUID to serve as a connection ID.
* Connection IDs serve as keys into a bag of ack properties. 
* The ack bag is a singleton and stored externally from events. It needs to be accessible from multiple threads, but Arcs are not serializable which would break the metric::Event enum traits.
* After sending raw events to the upstream sinks, the avro source waits until the acknowledge count becomes nonzero. The wait method is abstracted into the ack bag to allow future implementation changes w/o changing the semantics. 
* The wait method is currently a sleep-and-check loop with sane sleep values, however it may be possible to make it mio evented in a future change.

### DIFF
--- a/src/metric/ackbag.rs
+++ b/src/metric/ackbag.rs
@@ -1,3 +1,4 @@
+use std::cmp::min;
 use std::sync::{Arc, Mutex};
 use std::{thread, time};
 use util;
@@ -8,15 +9,17 @@ pub struct SyncAckProps {
     acks_received: usize
 }
 
-impl SyncAckProps {
-    pub fn new() -> Self {
+impl Default for SyncAckProps {
+    fn default() -> SyncAckProps {
         SyncAckProps {
             acks_received: 0,
         }
     }
+}
 
+impl SyncAckProps {
     pub fn ack(&mut self) {
-        self.acks_received.saturating_add(1);
+        self.acks_received = self.acks_received.saturating_add(1);
     }
 
     pub fn acks_received(&self) -> usize {
@@ -30,15 +33,16 @@ pub struct SyncAckBag {
     waiting_syncs: Arc<Mutex<SyncAckBagMap>>
 }
 
-
-impl SyncAckBag {
-    /// Creates a SyncAckBag
-    pub fn new() -> Self {
+impl Default for SyncAckBag {
+    fn default() -> SyncAckBag {
         SyncAckBag {
             waiting_syncs: Arc::new(Mutex::new(SyncAckBagMap::default()))
         }
     }
+}
 
+
+impl SyncAckBag {
     /// Return a clone of the SyncAckProps for the given key
     pub fn props_for(&self, key: Uuid) -> Option<SyncAckProps> {
         let bag = self.waiting_syncs.lock().unwrap();
@@ -51,7 +55,7 @@ impl SyncAckBag {
     /// Insert an empty-initialized SyncAckProps for the given key
     pub fn prepare_wait(&self, key: Uuid) {
         let mut bag = self.waiting_syncs.lock().unwrap();
-        bag.insert(key, SyncAckProps::new());
+        bag.insert(key, SyncAckProps::default());
     }
 
     /// Remove the given key from the internal bag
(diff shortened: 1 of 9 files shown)


src/metric/ackbag.rs
```diff
@@ -0,0 +1,104 @@
...
+        let mut current_nap_time = time::Duration::from_millis(5);
+        let mut wait = true;
+        while wait {
+            match self.props_for(key) {
+                Some(props) => wait = props.acks_received() == 0,
+                _ => wait = false
+            }
+            if wait {
+                thread::sleep(current_nap_time);
+                current_nap_time = current_nap_time * 2;
```
pulltab: Why not use a clamp method?  Oh, because there isn't one in the standard lib.

blt: One should be added to `time`, like: https://github.com/postmates/cernan/blob/master/src/time.rs#L42

There's also crates that are for this purpose. https://github.com/yoshuawuyts/exponential-backoff/ for example.

src/metric/ackbag.rs
```diff
@@ -0,0 +1,104 @@
...
+type SyncAckBagMap = util::HashMap<Uuid, SyncAckProps>;
+
+pub struct SyncAckBag {
+    waiting_syncs: Arc<Mutex<SyncAckBagMap>>
+}
+
+
+impl SyncAckBag {
+    /// Creates a SyncAckBag
+    pub fn new() -> Self {
```
blt: Recommend implementing Default for SyncAcBag versus having a blank new constructor.

src/metric/ackbag.rs
```diff
@@ -0,0 +1,104 @@
...
+use util;
+use uuid::Uuid;
+
+#[derive(Clone)]
+pub struct SyncAckProps {
+    acks_received: usize
+}
+
+impl SyncAckProps {
+    pub fn new() -> Self {
```
blt: Recommend implementing Default for SyncAckProps versus having a blank new constructor.

src/metric/ackbag.rs
```diff
@@ -0,0 +1,104 @@
...
+        SyncAckProps {
+            acks_received: 0,
+        }
+    }
+
+    pub fn ack(&mut self) {
+        self.acks_received.saturating_add(1);
```
blt: This function could be `&self`. Or do you mean to do `self.acks_received = self.acks_received.saturating_add(1);` ?

src/metric/ackbag.rs
```diff
@@ -0,0 +1,104 @@
+use std::sync::{Arc, Mutex};
```
blt: This module needs some tests.

dparton: Done

src/source/avro.rs
```diff
@@ -217,19 +220,28 @@ impl AvroStreamHandler {
         &mut self,
         mut chans: util::Channel,
         payload: Vec<u8>,
+        connection_id: Uuid
     ) -> Result<Option<u64>, PayloadErr> {
         match payload.into() {
             Payload::Valid { header, avro_blob } => {
+                let ackbag = metric::global_ack_bag();
```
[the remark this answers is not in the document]
dparton: I thought about where to put this and the reasons I chose metric are as follows:
1. The ack bag is logically coupled to metrics (currently just the Raw variant, but that is by use case, not by design limitation). It is a place to store information that a metric has been successfully ushered all the way to the final resting place.
2. The ack bag is *functionally* coupled to both sources and sinks. No individual source or sink is specifically responsible -- they're all just potential users. In this, avro->kafka just happens to be the first place that needs such a concept.

The usage of the ack bag comes up whenever a metric needs to flow from a source to a sink in a synchronous manner. That is, if a source should pause its work and wait for the sink to do the work requested and then resume once the sink has done it, the ack bag is used.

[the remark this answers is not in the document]
blt: Heard, makes good sense.

dparton:
Note: I had to repush the commits after merging #449 (required a rebase). Merging this w/o reapproval now that tests have completed again.

PR #450, opened 2018-09-10
fix(sink/influxdb): update influxdb port to 8086

goller:
Hey there, I'm a new rustacean but have been working for influx for a bit and have been looking around at what rust programs there are using the db.  cernan is cool !

I thought I'd send in a tiny patch;  This just corrects InfluxDB's default port to be  8086.


blt:
Looks great, thank you!

PR #451, opened 2019-01-15
Avro source protocol v2

josephglanville:
This introduces a new wire format version for the Avro source.

The existing version 1 Avro source doesn't provide any support
for extensible metadata.

Introduced in version 2 is a set of key value pairs in the header that
can be used to carry arbitrary metadata.

The protocol allows up to 255 key value pairs, each with a
key of up to 255 UTF-8 encoded bytes and a value of up to 65535 bytes in
length.

The new packet structure is described below in IETF RFC format:

```
 0                   1                   2                   3
 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                             Length                            |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                            Version                            |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                            Control                            |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                                                               |
+                               ID                              +
|                                                               |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                                                               |
+                            ShardBy                            +
|                                                               |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|   #KV Pairs   |   Key Length  |    Key (up to 255 bytes)      |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|          Value Length         |   Value (up to 65535 bytes)   |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                          Avro N Bytes                         |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

The source remains backwards compatible with version 1 clients.

### DIFF
--- a/src/metric/event.rs
+++ b/src/metric/event.rs
@@ -1,4 +1,5 @@
 use crate::metric::{LogLine, Telemetry};
+use crate::util::HashMap;
 use uuid::Uuid;
 
 /// Supported event encodings.
@@ -11,6 +12,9 @@ pub enum Encoding {
     /// JSON
     JSON,
 }
+
+/// Metadata: additional data attached to an event
+pub type Metadata = HashMap<Vec<u8>, Vec<u8>>;
 
 /// Event: the central cernan datastructure
 ///
@@ -41,6 +45,8 @@ pub enum Event {
         encoding: Encoding,
         /// Encoded payload.
         bytes: Vec<u8>,
+        /// Metadata used by some sinks
+        metadata: Option<Metadata>,
         /// Connection ID of the source on which this raw event was received
         connection_id: Option<Uuid>,
     },
--- a/src/sink/kafka.rs
+++ b/src/sink/kafka.rs
@@ -1,5 +1,5 @@
 //! Kafka sink for Raw events.
-use crate::metric;
+use crate::metric::{global_ack_bag, Encoding, Metadata};
 use crate::sink::Sink;
 use crate::source;
 use crate::util::Valve;
@@ -7,11 +7,12 @@ use futures::future::Future;
 use rdkafka::client::ClientContext;
 use rdkafka::config::{ClientConfig, RDKafkaLogLevel};
 use rdkafka::error::{KafkaError, RDKafkaError};
-use rdkafka::message::{Message, OwnedMessage};
+use rdkafka::message::{Headers, Message, OwnedHeaders, OwnedMessage};
 use rdkafka::producer::future_producer::DeliveryFuture;
-use rdkafka::producer::FutureProducer;
+use rdkafka::producer::{FutureProducer, FutureRecord};
 use rdkafka::util::current_time_millis;
 use std::collections::HashMap;
+use std::str;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use uuid::Uuid;
 
@@ -131,7 +132,7 @@ impl KafkaPublishable for KafkaPublishResult {
     }
 }
 
-type BoxedKafkaPublishable = Box<KafkaPublishable + Send + Sync>;
+type BoxedKafkaPublishable = Box<KafkaPublishable + Send>;
 
 trait KafkaMessageSender {
     fn try_payload(
@@ -139,6 +140,7 @@ trait KafkaMessageSender {
         topic: &str,
         payload: &[u8],
         key: &[u8],
+        metadata: Option<Metadata>,
         connection_id: Option<Uuid>,
     ) -> BoxedKafkaPublishable;
 }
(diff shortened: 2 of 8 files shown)


src/sink/kafka.rs
```diff
@@ -131,14 +132,15 @@ impl KafkaPublishable for KafkaPublishResult {
     }
 }
 
-type BoxedKafkaPublishable = Box<KafkaPublishable + Send + Sync>;
```
pulltab: Why is it safe to get rid of Sync here?

src/sink/kafka.rs
```diff
@@ -149,16 +151,21 @@ impl KafkaMessageSender for FutureProducer<STFUContext> {
...
         connection_id: Option<Uuid>
     ) -> BoxedKafkaPublishable {
+        let owned_headers = OwnedHeaders::new();
+        for (k, v) in headers {
+            owned_headers.clone().add(str::from_utf8(&k).unwrap(), &v);
+        };
         Box::new(KafkaPublishResult {
-            inner: Some(self.send_copy(
-                topic,
-                /* partition */ None,
```
pulltab: This seems lost in translation.  The new code will publish to a random partition?

src/source/avro.rs
```diff
@@ -82,6 +83,22 @@ impl From<Vec<u8>> for Payload {
         let mut cursor = Cursor::new(vec);
         let header: Header = (&mut cursor).into();
 
+        let mut key_pairs = HashMap::new();
+        // Read key pairs from wire if version > 1
```
pulltab: We will need to update the wiki documentation after this is merged.  See - https://github.com/postmates/cernan/wiki/SourcesAvro

src/metric/event.rs
```diff
@@ -41,6 +43,8 @@ pub enum Event {
         encoding: Encoding,
         /// Encoded payload.
         bytes: Vec<u8>,
+        /// Header values used by some sinks to send metadata
+        headers: HashMap<Vec<u8>, Vec<u8>>,
```
josephglanville: Could make `Option<HashMap<Vec<u8>, Vec<u8>>>`, probably worth it to avoid allocations.

src/source/avro.rs
```diff
@@ -324,16 +343,85 @@ mod test {
...
                     };
-                    assert!(payload == expected);
+                    assert_eq!(payload, expected);
                 }
             }
         }
     }
+
+    #[test]
+    fn parse_payload_happy_path_v2() {
```
josephglanville: Could DRY the tests up a bit.

pulltab:
@blt Still labelled a WIP, but can we have your eyes on this changeset too?

src/metric/event.rs
pulltab: I am 9/10s sure @blt would call you out on the allocations bit. :)

src/sink/kafka.rs
josephglanville: `Sync` allows types to be shared my multiple threads, `Send` means it can be sent between threads. In our case we don't need to use it again after sending to a different thread so `Send` is fine. Otherwise we would need to make the rdkafka `OwnedHeaders` `Sync` somehow and I didn't know how to do that so this seemed the easiest solution.

josephglanville:
@blt this is ready for initial review but is yet to undergo full end to end testing.

src/metric/event.rs
```diff
@@ -41,6 +42,8 @@ pub enum Event {
         encoding: Encoding,
         /// Encoded payload.
         bytes: Vec<u8>,
+        /// Header values used by some sinks to send metadata
```
blt: There are a lot of `Event` instances that get made inside cernan. I can't recall off the top of my head how the enum field variants we see here affect size. That is, does `Raw` contribute more to the total size of `Event` than `Telemetry(Telemetry)`? I can't recall off the top of my head. Worth doing a size_of to find out. 

The smaller `Event` is the better.

src/metric/event.rs
```diff
@@ -1,4 +1,5 @@
 use crate::metric::{LogLine, Telemetry};
+use std::collections::HashMap;
```
blt: Please prefer the use of `util::HashMap`.

src/metric/event.rs
```diff
@@ -41,6 +42,8 @@ pub enum Event {
         encoding: Encoding,
         /// Encoded payload.
         bytes: Vec<u8>,
+        /// Header values used by some sinks to send metadata
+        headers: Option<HashMap<Vec<u8>, Vec<u8>>>,
```
blt: You've had to duplicate `HashMap<Vec<u8>, Vec<u8>>` through the codebase, going to be a real pain if you decide to change any of the internal types. How about making a type alias for this?

josephglanville:
I type aliased `util::HashMap<Vec<u8>, Vec<u8>>` to `Metadata` which cleaned things up a bit. Though `Default:: default ()` feels a little janky. Would be nice if `Metadata::new()` worked without additional code. Might research that a bit if I get time.

josephglanville:
@blt fixed a bug in the header handling and improved test coverage to meet the targets.

PR #458, opened 2019-01-25
Make gzip compression optional in Prometheus handler

### LINKED ISSUE #456: Prometheus Sink - Make gzip Compression Optional
On a system which is cpu bound and exposes a large number of metrics, the usage of gzip in the prometheus sink results in undesired CPU spikes.

We should make gzip encoding a config option.  This will save the cpu overhead, making the prometheus sync more IO bound.

josephglanville:
Cernan can sometimes experience CPU exhaustion on very busy systems. This change is intended to reduce CPU requirements of the Prometheus handler to increase responsiveness on these systems.

In addition add some tests for the Prometheus sink to increase test coverage.

Fixes #456

### DIFF
--- a/src/sink/influxdb.rs
+++ b/src/sink/influxdb.rs
@@ -424,10 +424,10 @@ mod test {
         );
         let mut buffer = String::new();
         influxdb.format_stats(&mut buffer, influxdb.aggr_slice());
-        let lines: Vec<&str> = buffer.lines().collect();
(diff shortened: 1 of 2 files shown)


blt:
@josephglanville cool, good work. The benchmarks are a good start. The trick with benching cernan has always been generating enough load to match the extreme conditions folks report, which hampers the benchmark since they're co-local. I would not mind seeing what you have in `jpg/prom-gzip-bench` merged up to master.

